PR #2, opened 2023-01-20
Changes query properties type to Double to support fractional values

Saik0s:
Change the data type of `temperature`, `top_p`, `frequency_penalty`, and `presence_penalty` to `Double` to allow for fractional values.


Krivoblotsky:
Thanks for the contribution, @Saik0s!
I'm wondering how could I miss those types:)

PR #3, opened 2023-02-07
Add timeout to let user specify the timeout interval for the quest

yuanhang:
Some requests will take longer than others, and this PR allows developers to specify the appropriate request timeout length for the situation, rather than using the default 60 seconds for all requests.


Krivoblotsky (approved):
Hey, @yuanhang!
Thanks for your contribution!

PR #5, opened 2023-02-11
Keep CompletionQuery struct up to date

Janldeboer:
The old snippet was not up to date / not like the actual code base. I was surprised that most values weren't optional when in fact they were.


Krivoblotsky (approved):
Thanks, @Janldeboer!

PR #15, opened 2023-03-10
Add enum `Chat.Role`…

tomdai:
Add enum `Chat.Role` so that `Chat` has a safer initializer 

```Swift
init(role: Role, content: String)
```

than

```Swift
init(role: String, content: String)
```

### DIFF
--- a/Sources/OpenAI/OpenAI.swift
+++ b/Sources/OpenAI/OpenAI.swift
@@ -147,9 +147,19 @@ public extension OpenAI {
         public let role: String
         public let content: String
         
+        public enum Role: String {
+            case system
+            case assistant
+            case user
+        }
+        
         public init(role: String, content: String) {
             self.role = role
             self.content = content
+        }
+        
+        public init(role: Role, content: String) {
+            self.init(role: role.rawValue, content: content)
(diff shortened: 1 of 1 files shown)


Krivoblotsky:
Hey, @tomdai!
Thanks for the addition.

Krivoblotsky (approved):

alelordelo:
Hi @tomdai,

Did you  get the Chat.Role to work?

I got this bug:
https://github.com/MacPaw/OpenAI/issues/98

tomdai:
Hi @alelordelo,

`Chat.Role` did work at the time I made this PR. The implementation of `Chat.Role` has changed a lot since then though.

---

Screenshot 2023-09-21 at 18 01 09@2x

I did try to reproduce the issue #98, however it looks like the message created after sending the prompt doesn't have the role `.function` but `.assistant` instead. 

So I don't think the issue is with the implementation of `Chat.Role`.

alelordelo:
It works with simulator, but not in built app. The reason is (as you pointed) "the message created after sending the prompt doesn't have the role .function but .assistant instead"

PR #19, opened 2023-03-18
Update Readme and replace deprecated .branch("main")

MSWagner:
The usage of .branch("main") is deprecated and should be replaced with branch: "main"


Krivoblotsky (approved):
Thanks!

PR #38, opened 2023-04-08
Add support for the Models API

SunburstEnzo:
Hi there! This is my first pull request to the project so I'm sure I've missed something or not quite done something in the theme of things, so do please let me know and I will make the required changes as soon as possible.

## What

• Added ModelQuery/Result and ModelsQuery/Result (ModelsQuery has no input parameters)
• Added ModelType to avoid conflicting with Models.swift (which could be renamed ModelIdentifier maybe?)
• Moved all Model related files into Models subfolder (inside Models, could be confusing however)
• Added withPath function to allow passing in a path parameter to APIPath (as required with the Models API)
• Unit tests have been added

Permissions was not added at this stage, but could come later. The property is also left vague in the documentation.

## Why

Obtaining the list of models allows for more flexibility and retrieving model information can help with understanding ownership.

## Affected Areas

Models.swift file location moved, otherwise only enhancements.


Krivoblotsky (approved):
Hey, @SunburstEnzo!
Thanks for the addition, it looks perfect to me 👍

PR #40, opened 2023-04-09
ModelResult typo fix along with others

SunburstEnzo:
I was going through my previous Model PR merge and spotted ModelsResult was being used instead of ModelResult in just one instance, but then spotted other things, so hopefully a "typo fix" PR makes sense in that context! Let me know if not and I'll amend.

## What

• ModelResult instead of ModelsResult in one instance (knew there was something!)
• "repositorty" → "repository"
• "MutlipartFormDataBodyBuilder" → "MultipartFormDataBodyBuilder"
• `func append` indentation tweaked for consistency
• "transriptions" → "transcriptions"
• "reqest" → "request"

## Why

Should help with consistency and readability.

## Affected Areas

Readme, OpenAI.swift, Audio, Tests.


Krivoblotsky (approved):
Thanks for catching 👍

Issue #41, opened 2023-04-10
Add Decoder Tests


SunburstEnzo:
**Is your feature request related to a problem? Please describe.**
I've discovered an issue with ModelResult by having a unit test.

**Describe the solution you'd like**
Adding a set of decoder-related unit tests would ensure the models are correct when comparing with the expected json.

**Describe alternatives you've considered**
It would be handy to have examples in the sample project, but unit tests would still be the most optimum solution.

**Additional context**
My proposition is something similar to the below, but with varying examples for all supported endpoint results

`class OpenAITestsDecoder: XCTestCase`
        
    func testModelsSingle() async throws {
        let data = """
        {
          "data": [
            {
              "id": "gpt-3.5-turbo",
              "object": "model",
              "owned_by": "organization-owner"
            }
          ],
          "object": "list"
        }
        """.data(using: .utf8)!
        
        do {
            let decoded = try JSONDecoder().decode(ModelsResult.self, from: data)
            
            let mock = ModelsResult(data: [
                .init(id: .gpt3_5Turbo, object: "model", ownedBy: "organization-owner")
            ], object: "list")
            
            XCTAssertEqual(decoded, mock)
            
        } catch let error {
            XCTFail(error.localizedDescription)
        }
    }

Krivoblotsky:
Thanks for your contribution, @SunburstEnzo!

PR #42, opened 2023-04-10
Add OpenAITestsDecoder and removed ModelType

SunburstEnzo:
## What

• Add OpenAITestsDecoder.swift with JSONDecoder related (very simple equatable) tests for all results with example json. #41 
• Removed ModelType as it was deemed unnecessary.

## Why

Adding a unit test for ModelResult showed that I'd misread the json, so adding unit tests for all would help make sure no other mistakes were made with the Codable structures. It also ensures compatibility with updates to the API from OpenAI.

## Affected Areas

ModelResult, ModelsResult, expanding on current results will need the relevant unit test tweaks


Tests/OpenAITests/OpenAITestsDecoder.swift
```diff
@@ -0,0 +1,205 @@
...
+@available(tvOS 13.0, *)
+class OpenAITestsDecoder: XCTestCase {
+    
+    private let decoder = JSONDecoder()
+    
+    override func setUp() {
+        super.setUp()
+    }
+    
+    func testCompletions() async throws {
```
Krivoblotsky, line 22: What do you think about some simplification by moving common code to the function: 

```swift
class OpenAITestsDecoder: XCTestCase {

    private func decode<T: Decodable>(_ jsonString: String, _ expectedValue: T) throws {
        let decoder = JSONDecoder()
        let data = jsonString.data(using: .utf8)!
        let decoded = try decoder.decode(T.self, from: data)
        XCTAssertEqual(decoded, expectedValue)
    }

    func testCompletions() throws {
        let data = """
        {
          "id": "foo",
          "object": "text_completion",
          "created": 1589478378,
          "model": "text-davinci-003",
          "choices": [
            {
              "text": "Hello, world!",
              "index": 0,
              "logprobs": null,
              "finish_reason": "length"
            }
          ],
          "usage": {
            "prompt_tokens": 5,
            "completion_tokens": 7,
            "total_tokens": 12
          }
        }
        """
        let expectedValue = CompletionsResult(id: "foo", object: "text_completion", created: 1589478378, model: .textDavinci_003, choices: [.init(text: "Hello, world!", index: 0)], usage: .init(promptTokens: 5, completionTokens: 7, totalTokens: 12))
        try decode(data, expectedValue)
    }

   //Other tests here
}
```

SunburstEnzo:
Thanks for the suggestion @Krivoblotsky, I've added your suggestion while also ensuring the type conforms to Equatable as well. Do let me know if you'd like any additional changes 😊

Krivoblotsky (approved):

Krivoblotsky:
Thanks!

PR #55, opened 2023-04-18
Add missing public init to ModelsQuery and ModelQuery

jongyeol:
## What

Add missing public initializers to `ModelsQuery` and `ModelQuery`.

## Why

Previously, there were no public initializers of `ModelsQuery` and `ModelQuery`. So, users of this library cannot use `models()` and `model()`.

```swift
let openAI = OpenAI(apiToken: "...")
let query = ModelsQuery() // Missing argument for parameter 'from' in call
try await openAI.models(query: query)
```

## Affected Areas

Now, we can use `models()` and `model()` properly.


DJBen (approved):

Krivoblotsky (approved):
Thanks!

PR #56, opened 2023-04-18
Use GET for model and models

jongyeol:
## What

`/v1/models` and `/v1/models/{model}` should use `GET` method instead of `POST`.

## Why

https://platform.openai.com/docs/api-reference/models/list

## Affected Areas

Now, `model()` and `models()` works.


jongyeol:
@Krivoblotsky There was one more thing. Please take a look when you have time.

Sources/OpenAI/Private/JSONRequest.swift
```diff
@@ -33,7 +33,9 @@ extension JSONRequest: URLRequestBuildable {
             request.setValue(organizationIdentifier, forHTTPHeaderField: "OpenAI-Organization")
         }
         request.httpMethod = method
-        request.httpBody = try JSONEncoder().encode(body)
+        if method != "GET" && method != "DELETE" {
```
Krivoblotsky, line 36: What do you think of extracting this condition into a function or variable? Something like `requiresBody: Bool`?

Thanks for contributing, really appreciate it 👍

jongyeol: SGTM 😄 
I've updated the PR to use the optional body instead.
On the other side, `ModelsQuery` itself is not necessary too. Do you want to keep it?

SunburstEnzo:
Good catch! Models was my first PR and I knew I missed a bunch of stuff 🤦‍♂️

Sources/OpenAI/Private/JSONRequest.swift
Krivoblotsky: Yes, looks like `ModelsQuery` is unused. Do you mind removing it with this PR? Along with `models` function update. Looks like YAGNI.

Krivoblotsky:
@SunburstEnzo, that's ok) This is exactly what pull requests and code reviews exist for 👌

Sources/OpenAI/Private/JSONRequest.swift
jongyeol: No problem. Added a commit to remove `ModelsQuery` with updating the `models()` function.

Krivoblotsky (approved):

Issue #14, opened 2023-03-08
Streaming is broken


randomsnowflake:
**Code**
```swift
import Foundation
import OpenAI

let openAI = OpenAI(apiToken: "sk-...")

let query = OpenAI.ChatQuery(model: .gpt3_5Turbo, messages: [.init(role: "user", content: "hi!")], stream: true)
let result = try await openAI.chats(query: query)
```

**Exception**
```
Swift/ErrorType.swift:200: Fatal error: Error raised at top level: Swift.DecodingError.dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Error Domain=NSCocoaErrorDomain Code=3840 "Invalid value around line 1, column 0." UserInfo={NSDebugDescription=Invalid value around line 1, column 0., NSJSONSerializationErrorIndex=0})))

2023-03-08 14:37:01.246548+0100 CLIChat[81308:5267103] Swift/ErrorType.swift:200: Fatal error: Error raised at top level: Swift.DecodingError.dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Error Domain=NSCocoaErrorDomain Code=3840 "Invalid value around line 1, column 0." UserInfo={NSDebugDescription=Invalid value around line 1, column 0., NSJSONSerializationErrorIndex=0})))
```

It works with `stream: false`

Krivoblotsky:
Hey, @randomsnowflake.
Thanks for bringing this to our attention. 
Looks like streaming is unsupported for now and we need to implement server-sent events for this.

longseespace:
I hope this will be implemented soon. Meanwhile, you can use this alternative (only for streaming - I'm using both clients in my app)

https://github.com/nate-parrott/openai-streaming-completions-swift

randomsnowflake:
I already use it. I think it's also the only Swift OpenAI API with streaming support. 

However, have a look at my issue report here: [Missing Model Parameters and No Error Message](https://github.com/nate-parrott/openai-streaming-completions-swift/issues/3) ... it's not as mature as I would like. Looking forward to MacPaw's implementation of streaming.

Krivoblotsky:
I know that @rshimoda is on his way to adding streaming support here, so I believe it must be implemented shorty 👌

DJBen:
[...]

PR #57, opened 2023-04-18
Add streaming session and ability to use streaming

Krivoblotsky:
## What

1. Chats and Completions have been updated with streaming functionality
2. ChatStreamingResult has been created
3. Tests have been updated

## Why

#14 

## Affected Areas

Completions, Chats


Sources/OpenAI/OpenAI.swift
```diff
@@ -35,6 +35,7 @@ final public class OpenAI: OpenAIProtocol {
     }
     
     private let session: URLSessionProtocol
+    private var streamingSessions: [NSObject] = []
```
DJBen, line 38: You may want [`NSHashTable`](https://developer.apple.com/documentation/foundation/nshashtable) to give it set semantics and broader memory operations.

Sources/OpenAI/OpenAI.swift
```diff
@@ -64,7 +65,11 @@ final public class OpenAI: OpenAIProtocol {
     }
     
     public func chats(query: ChatQuery, completion: @escaping (Result<ChatResult, Error>) -> Void) {
-        performRequest(request: JSONRequest<ChatResult>(body: query, url: buildURL(path: .chats)), completion: completion)
+        if query.stream == true {
```
DJBen, line 68: nit: query.stream

Sources/OpenAI/Public/Models/ChatResult.swift
```diff
@@ -10,13 +10,20 @@ import Foundation
...
     public struct Choice: Codable, Equatable {
+        public struct Delta: Codable, Equatable {
+            public let content: String?
+            public let role: Chat.Role?
+        }
+        
         public let index: Int
-        public let message: Chat
-        public let finishReason: String
+        public let message: Chat?
```
DJBen, line 19: If you want to mix the complete and partial results, can you comment in what condition these properties are `nil`?

I have moderate preference in separating a complete `Choice` from a `PartialChoice`, provided that they have major differences, despite the official OpenAI API mix them together.

```swift
// Choice stays the same

public struct PartialChoice: Codable, Equatable {
  public let index: Int
  public let delta: Delta
}

```

DJBen:
Since we are using two separate functions for the `chats(query:)` and `chatStream(query:)` to represent streaming and non-streaming version of the chat api, we can potentially get rid of the `stream: Bool?` property inside `ChatQuery`, and modify the encoder to serialize `stream: false` and `stream: true` respectively when calling these two different functions. But it's probably a large model change.

DJBen:
I added streaming in the demo app #59

DJBen (approved):

Krivoblotsky:
It looks like everything is working as expected 🤞 
Going to check more tests on weekends and proceed. 
https://user-images.githubusercontent.com/1411778/235212448-0aa844d5-7ffb-487e-a0bb-37eed5666f48.mov

Sources/OpenAI/OpenAI.swift
```diff
@@ -127,7 +132,26 @@ extension OpenAI {
...
+            let request = try request.build(token: configuration.token, organizationIdentifier: configuration.organizationIdentifier, timeoutInterval: configuration.timeoutInterval)
+            let session = StreamingSession<ResultType>(urlRequest: request)
+            session.onReceiveContent = {_, object in
+                completion(.success(object))
+            }
+            session.onProcessingError = {_, error in
+                completion(.failure(error))
+            }
+            session.onComplete = { [weak self] object, error in
+                self?.streamingSessions.removeAll(where: { $0 == object })
```
tikimcfee, line 149: Heya! Jumping in, I got a little excited to test this out. Doing some testing myself, this needs to call `completion` in some way in order for the backing `AsyncStream` to actually finish. At the moment I'm throwing an error like this:
```
enum StreamError: Error {
 case streamStop
}
//...
completion(.failure(StreamError.streamStop)
```

Which allows the `throw` to act as the trigger to stop. There's likely a more elegant way to do this, so hopefully the report at least helps track that part down. Cheers, and thanks much for your time!

DJBen:
Hi @Krivoblotsky do you know when this can be merged?

Krivoblotsky:
Hi, @DJBen 
I have plans for this week for this to be updated and merged.

Sources/OpenAI/Public/Models/ChatResult.swift
Krivoblotsky: I've decided to keep `ChatResult`to not break backward compatibility and created "ChatStreamingResult" along with a separate function for streaming. I found this solution more straightforward and clean. What do you think about this?

Krivoblotsky:
Hey, @DJBen. 
I'm ready to merge it, but wanted to ask you for a review first 🙏

DJBen (approved):
LG!

PR #59, opened 2023-04-19
Adopt streaming in Demo app

DJBen:
## What

Adopt streaming in demo app, and add a model switcher between gpt3 and 4 models.

https://user-images.githubusercontent.com/2078653/233036013-aa339689-e2d3-40db-a879-1197b478be70.mov

## Why

Demo app will be broken if this doesn't come in. Also I think this is how people want to use the API primarily.

## Affected Areas
Demo app.

### DIFF
--- a/Demo/DemoChat/Sources/ChatStore.swift
+++ b/Demo/DemoChat/Sources/ChatStore.swift
@@ -53,17 +53,27 @@ public final class ChatStore: ObservableObject {
     }
     
     @MainActor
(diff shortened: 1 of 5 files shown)


rotemdoron:
any plan to include this?

DJBen:
@rotemdoron I do. Ultimately I defer to @Krivoblotsky

Krivoblotsky:
I’m ok to merge it into the streaming branch, however, I’m gonna run more tests before merging streaming branch itself to handle some corner cases.

Issue #46, opened 2023-04-11
Add images/create-variation


Krivoblotsky:
Extend the library by adding images/create-variation capabilities

### Documentation
https://platform.openai.com/docs/api-reference/images/create-variation

ingvarus-bc:
Tested it out, seems to be working fine. So closing this issue ❤️

Issue #45, opened 2023-04-11
Add image/edit


Krivoblotsky:
Extend the library by adding image/edit capabilities

### Documentation
https://platform.openai.com/docs/api-reference/images/create-edit

ingvarus-bc:
@Krivoblotsky there was no action taken to your:
> please remove the comment

I removed it, everything seems to work properly, I see an approve from you, closing the issue and merging the [PR-62](https://github.com/MacPaw/OpenAI/pull/62)

PR #62, opened 2023-04-24
Extend Images API to support Edits and Variations

SunburstEnzo:
## What

(I wasn't sure about when it should be called Image or Images, Edit or Edits, etc. but I think I was consistent at least!)

• Added ImageEditsQuery and ImageVariationsQuery (both still use ImagesResult as the response)
• Removed mention of model as a parameter for ImagesQuery in readme as that isn't currently supported

## Why

More image generation abilities
Edit: #45 and #46 

## Affected Areas

Did not alter ImagesQuery but may be best to tweak the wording for consistency? Maybe they should all be "CreateImagesQuery", "CreateImageEditQuery", etc. – not sure

I'm not currently in a situation to use ChatGPT Plus so can't fully test

Happy to make any tweaks 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ This repository contains Swift community-maintained implementation over [OpenAI]
     - [Completions](#completions)
     - [Chats](#chats)
     - [Images](#images)
+        - [Create Image](#create-image)
+        - [Create Image Edit](#create-image-edit)
+        - [Create Image Variation](#create-image-variation)
     - [Audio](#audio)
         - [Audio Transcriptions](#audio-transcriptions)
         - [Audio Translations](#audio-translations)
@@ -252,6 +255,8 @@ Given a prompt and/or an input image, the model will generate a new image.
 
 As Artificial Intelligence continues to develop, so too does the intriguing concept of Dall-E. Developed by OpenAI, a research lab for artificial intelligence purposes, Dall-E has been classified as an AI system that can generate images based on descriptions provided by humans. With its potential applications spanning from animation and illustration to design and engineering - not to mention the endless possibilities in between - it's easy to see why there is such excitement over this new technology.
 
+### Create Image
+
 **Request**
 
 ```swift
@@ -276,6 +281,7 @@ struct ImagesResult: Codable, Equatable {
     public let data: [URLResult]
 }
 ```
+
 **Example**
 
 ```swift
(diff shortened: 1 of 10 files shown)


SunburstEnzo:
I was thinking, maybe it's best to keep one images() function and pass in an enum with the query as an associated value?

Something like
`enum ImageQueryType {
    case create(ImagesQuery)
    case edit(ImageEditsQuery)
    case variation(ImageVariationsQuery)
}`

`public func images(queryType: .create(ImagesQuery(), completion: @escaping (Result<ImagesResult, Error>) -> Void)`

README.md
```diff
@@ -300,6 +306,74 @@ let result = try await openAI.images(query: query)
...
+### Create Image Edit
+
+Creates an edited or extended image given an original image and a prompt.
+
+**Request**
+
+```swift
+public struct ImageEditsQuery: Codable {
+    /// The image to edit. Must be a valid PNG file, less than 4MB, and square. If mask is not provided, image must have transparency, which will be used as the mask.
+    public let image: String
```
Krivoblotsky, line 318: Could you describe why the image has a `String` type? Looks like, this should be a multipart form data request where the image and mask should be uploaded.

README.md
```diff
@@ -300,6 +306,74 @@ let result = try await openAI.images(query: query)
...
+    /// An additional image whose fully transparent areas (e.g. where alpha is zero) indicate where image should be edited. Must be a valid PNG file, less than 4MB, and have the same dimensions as image.
+    public let mask: String?
```
Krivoblotsky, line 320: please see the comment above

Krivoblotsky (changes requested):
Hey, @SunburstEnzo 
Thanks for the update. 
Looks like we need to upload images, not just pass their names. 
The documentation is a bit odd https://platform.openai.com/docs/api-reference/images/create-edit

Do you have any working example to demonstrate the approach?

SunburstEnzo:
Yeah great points, I didn't fully understand it myself, so I was wondering if any image url would work or if it had to be uploaded using the Files API.

I'm now on the pay as you go OpenAI plan so I'll try and see what errors I get back with a valid api key and work backwards. 

May be best for me to add basic Files API support and come back?

SunburstEnzo:
Ok I think I've found out how to do it in part

First error was that it isn't json to be sent:
```
Invalid Content-Type header (application/json), expected multipart/form-data. (HINT: If you're using curl, you can pass -H 'Content-Type: multipart/form-data')
```

So that meant tweaking the header to:
```
let boundary = UUID().uuidString
request.setValue("multipart/form-data; boundary=\(boundary)", forHTTPHeaderField: "Content-Type")
```

Then it was figuring out the formatting of the httpBody to send (as NSMutableData):
```
'image' is a required property
```

I found a StackOverflow article that did pretty much what I wanted (https://stackoverflow.com/a/47571172/1241153):
```
let body = NSMutableData()
let fname = "whitecat.png"

let mimetype = "image/png"
//define the data post parameter
body.append("--\(boundary)\r\n".data(using: String.Encoding.utf8)!)
body.append("Content-Disposition:form-data; name=\"image\"\r\n\r\n".data(using: String.Encoding.utf8)!)
body.append("hi\r\n".data(using: String.Encoding.utf8)!)
body.append("--\(boundary)\r\n".data(using: String.Encoding.utf8)!)
body.append("Content-Disposition:form-data; name=\"image\"; filename=\"\(fname)\"\r\n".data(using: String.Encoding.utf8)!)
body.append("Content-Type: \(mimetype)\r\n\r\n".data(using: String.Encoding.utf8)!)
body.append(imageData)
body.append("\r\n".data(using: String.Encoding.utf8)!)
body.append("--\(boundary)--\r\n".data(using: String.Encoding.utf8)!)
request.httpBody = body as Data
```

And it worked with the hardcoded image request (using the whitecat image in the readme as the base image)!
Screenshot 2023-04-25 at 4 14 57 pm 2

I haven't figured out how to use the query to add the size and n properties in but I got a result:
img-J1YDYUniLdkjs8vEdZYan4Yu

Verified API usage with OpenAI/Dall-E
Screenshot 2023-04-25 at 4 16 51 pm

Krivoblotsky:
@SunburstEnzo, you can use `MultipartFormDataBodyEncodable` protocol for that. 
Please see `AudioTranslationQuery` as a reference.

SunburstEnzo:
🤦‍♂️ nice one! Didn't spot that, should've read more on what was already in the project. I'll try and get it working like that 👍

SunburstEnzo:
@Krivoblotsky I've updated `ImageEditsQuery` and `ImageVariationsQuery` to use `MultipartFormDataBodyEncodable` as mentioned  .

Had to edit the `MultipartFormDataEntry` enum to allow for optional filename and file when using a mask image, but instead I could add an if let just for the mask entry in the BodyBuilder if that’s more favourable.

 Also another white cat image but sized to 256x256 by Dall-E:
img-SD6nj7ztSNntVx2m7uARyhlS

Tests/OpenAITests/OpenAITests.swift
```diff
@@ -62,6 +62,45 @@ class OpenAITests: XCTestCase {
         XCTAssertEqual(inError, apiError)
     }
     
+    func testImageEdit() async throws {
+//        let imageData = Data()
```
Krivoblotsky, line 66: Please remove the comment

Krivoblotsky (approved):

ingvarus-bc (approved):

Issue #49, opened 2023-04-13
[Demo app] Add audio transcription example; add image generation example


DJBen:
**Is your feature request related to a problem? Please describe.**
The demo app includes a chat example; audio and image generation remains TBD.

**Describe the solution you'd like**
Using the OpenAI API, add audio transcribe and image generation demos.

**Describe alternatives you've considered**
Not doing them.

DJBen:
Would be nice if you can tag this as "help wanted"

Krivoblotsky:
Sure. Done ✅

SunburstEnzo:
I've added an Image Generation example to the demo – let me know if there's anything you'd like changed in it 😊 I think it could definitely be improved to _show_ the images rather than just list the URLs but didn't know what would be best to load those images (if we need a cache, etc – worth using something like [SDWebImage](https://github.com/SDWebImage/SDWebImageSwiftUI)?)

nezhyborets:
As @SunburstEnzo added Image Generation, I've renamed the issue to just "Audio Transcription" example to make it smaller and more focused

PR #63, opened 2023-04-24
Add Create Image example to Demo app

SunburstEnzo:
## What

• Added ImageStore, follows same structure as ChatStore and MiscStore
• Added ImageView which is a list of being able to Create Images (this PR), and upcoming Create Image Edit and Create Image Variation (disabled, placeholder views)
• Uses SafariViewController to display image URL's when tapped though not ideal as not SwiftUI friendly
• ImagesResult.URLResult conforms to Hashable so I could use it in a list – may be better to map the url strings instead

## Why

#49, most likely a popular feature

## Affected Areas

ContentView, ImagesResult.URLResult

Like my Models DemoApp example, needs better error messaging

More than happy to tweak, especially in favour of nicer SwiftUI usage 👌

Preview:
Simulator Screenshot - iPhone 14 Pro - 2023-04-24 at 15 34 59 2
Simulator Screenshot - iPhone 14 Pro - 2023-04-24 at 17 21 54 2

### DIFF
--- a/Demo/App/APIProvidedView.swift
+++ b/Demo/App/APIProvidedView.swift
@@ -12,6 +12,7 @@ import SwiftUI
 struct APIProvidedView: View {
     @Binding var apiKey: String
     @StateObject var chatStore: ChatStore
+    @StateObject var imageStore: ImageStore
     @StateObject var miscStore: MiscStore
     @State var isShowingAPIConfigModal: Bool = true
 
@@ -29,6 +30,11 @@ struct APIProvidedView: View {
                 idProvider: idProvider
             )
         )
+        self._imageStore = StateObject(
+            wrappedValue: ImageStore(
+                openAIClient: OpenAI(apiToken: apiKey.wrappedValue)
+            )
+        )
         self._miscStore = StateObject(
             wrappedValue: MiscStore(
                 openAIClient: OpenAI(apiToken: apiKey.wrappedValue)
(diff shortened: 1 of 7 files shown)


Krivoblotsky:
What do you think about having a chat-like interface, but with images as a result instead of texts? So you can type a prompt and get an image right in the chat?

SunburstEnzo:
Yeah that's a good idea, I definitely thought about that initially but thought it might be overkill for the demo app? I wasn't sure how large the bottom keyboard-related area would end up being though. It could be that the (non-text/prompt) options are all in one context menu.

I'll have a think and see what I can mock up.

Krivoblotsky:
We can refer to Apple's Messages app here, what do you think? @DJBen do you have any thoughts?
IMG_5026

SunburstEnzo:
We could use the Link Presentation framework to display the images like that too https://developer.apple.com/documentation/linkpresentation

SunburstEnzo:
Link Presentation isn't super friendly with SwiftUI but found a few resources online to get it working, so it is a viable option (just obviously then in the chat format rather than this table format)

Simulator Screenshot - iPhone 14 Pro - 2023-04-25 at 12 03 42 2

SunburstEnzo:
The options for chat-style interaction is my main question though, something like this? Or "…" icon?
For image editing a "photo album" icon next to "..." could be a good answer that then covers both use cases

Simulator Screenshot - iPhone 14 Pro - 2023-04-25 at 13 06 29 2

ingvarus-bc (approved):

Issue #76, opened 2023-06-14
Support new models (gpt-4-0613 and gpt-3.5-turbo-0613)


iTarek:
Hello
We need support for the new Function calling.

> Developers can now describe functions to gpt-4-0613 and gpt-3.5-turbo-0613, and have the model intelligently choose to output a JSON object containing arguments to call those functions. This is a new way to more reliably connect GPT's capabilities with external tools and APIs.

[Function calling and other API updates](https://openai.com/blog/function-calling-and-other-api-updates)

ljoukov:
https://github.com/MacPaw/OpenAI/pull/77

ljoukov:
Implementation of functions support: https://github.com/MacPaw/OpenAI/pull/78

iTarek:
Thank you. When can we expect these features? I am very excited to try the new functions.

Krivoblotsky:
Merged, thanks @ljoukov 👍

PR #78, opened 2023-06-14
Add support for "functions" in OpenAI chat completion APIs.

ljoukov:
## What

Add support for "function calling" in chat completion API: https://platform.openai.com/docs/guides/gpt/function-calling

Chat.content type was changed from String to String? as it is now optional and is actually null for function calls.

## Why

OpenAI added support for function calling which need explicit API support.

## Affected Areas

Added:
ChatFunctionCall
ChatFunctionDeclaration

Updated:
ChatQuery (describes what function calls are available)
Chat (responses from OpenAI calling functions)

Added example to README.md


ljoukov:
This should close the issue: https://github.com/MacPaw/OpenAI/issues/76

ljoukov:
Switched to encoding JSONSchema as swift structs (as opposed to dictionaries or arbitrary values), code is not better type checked and much easier to read.

jakemor:
Tysm!

Krivoblotsky (approved):
hey, @ljoukov!
Thanks for the contribution!
Looks good to me, I'll do some tests in a while, and merge it.

fiatdv:
Seems like we need a bunch of good tests or some sample code of how to use functions in order to be truly usable?

ljoukov:
> Seems like we need a bunch of good tests or some sample code of how to use functions in order to be truly usable? @ljoukov can you pls add some additional support for functions? Trying to get the name of the function to be called but it's not working.

Hey @fiatdv did you check the example in README.md? It shows the response OpenAI gives you. Please make sure there is no misconception: the "functions" are not really functions are are not "called".

fiatdv:
@ljoukov I am trying to run the weather example as shown here: https://platform.openai.com/docs/guides/gpt/function-calling 
I am failing to see the response:

# Step 2, check if the model wants to call a function
    if message.get("function_call"):
        function_name = message["function_call"]["name"]
        function_args = json.loads(message["function_call"]["arguments"])

What I see is:

finishReason: Optional("function_call")

But no way to get the function_name or function_args because they are nil.

So, am wondering if this is working for you, or if you tried these steps. I know it is not going to call anything, but you should at least see these elements, right?

Maybe I am doing something wrong. Can you check? Thanks.

fiatdv:
Added a PR for testing purposes, here: https://github.com/ljoukov/OpenAI/pull/1
to support testing of this function. It would be great to have a way to define functions and provide callbacks to them when triggered in the chat.

ljoukov:
> Added a PR for testing purposes, here: [ljoukov#1](https://github.com/ljoukov/OpenAI/pull/1) to support testing of this function. It would be great to have a way to define functions and provide callbacks to them when triggered in the chat.

Great find! Indeed streaming was not working properly with functions. Also made JSONSchema members public so we no longer need to manually write JSON as strings and we can instead use strongly typed Swift structs:)
https://github.com/MacPaw/OpenAI/pull/78/commits/2620356f08bfa6ea6f5f3cc3618fc93cc71e57a6

fiatdv:
Thanks @ljoukov !! Works like a charm! Maybe as another PR we can add a mechanism to provide functions and call them when matched?

fiatdv:
Actually... looking at it closer, we should not have the Weather Code inside the ChatStore. Perhaps it should be injected into the store and when a function match is issued, the function would be called and the result would be pasted into the messageText? This is will make it more expandable... maybe we can have a FunctionStore with a Dictionary of function names and callbacks. Not sure if there is a better way @ljoukov ?

ljoukov:
> Actually... looking at it closer, we should not have the Weather Code inside the ChatStore. Perhaps it should be injected into the store and when a function match is issued, the function would be called and the result would be pasted into the messageText? This is will make it more expandable... maybe we can have a FunctionStore with a Dictionary of function names and callbacks. Not sure if there is a better way @ljoukov ?

IMHO the API part is already at the right level, do you agree?
The demo app of course might be improved and real function calls added, for example, to provide current time, so it is easy to implement. Again this is supposed to be a demo for how to use the APIs, we probably want to postpone improvements to a separate PR.

fiatdv:
Sure, agreed, many thanks for your contribution. The only thing, is please comment out the weather function and pass and empty functions array to avoid adding this to chat requests.Thanks again!Sent from my iPhoneOn Jun 16, 2023, at 6:28 PM, Liudmila Joukovskaya ***@***.***> wrote:﻿

Actually... looking at it closer, we should not have the Weather Code inside the ChatStore. Perhaps it should be injected into the store and when a function match is issued, the function would be called and the result would be pasted into the messageText? This is will make it more expandable... maybe we can have a FunctionStore with a Dictionary of function names and callbacks. Not sure if there is a better way @ljoukov ?

IMHO the API part is already at the right level, do you agree?
The demo app of course might be improved and real function calls added, for example, to provide current time, so it is easy to implement. Again this is supposed to be a demo for how to use the APIs, we probably want to postpone improvements to a separate PR.

ljoukov:
> Sure, agreed, many thanks for your contribution. The only thing, is please comment out the weather function and pass an empty functions array to avoid adding this to chat requests. Thanks again!

commenting out the code is likely not the best way to go. we should have a demo a function in a demo app, I can make it actually do something like current time so folks can use that. does this make sense?

fiatdv:
Yes, good idea! Thanks again.

pgorzelany:
Any chance this will get merge and released anytime soon?

PR #82, opened 2023-06-26
Add functionCall to ChatQuery

nathanborror:
ChatQuery was  missing `functionCall` which can be used to force the usage of given function names. API docs can be found here: https://platform.openai.com/docs/api-reference/chat/create#chat/create-function_call

It's a bit of a weird value type so I had to do some custom encoding.


Krivoblotsky (approved):
LGTM, thanks!

PR #89, opened 2023-07-12
Add initializer to ChatFunctionCall to initialize from external with properties

ysak-y:
## What

Add initializer with properties to `ChatFunctionCall` struct.

## Why

Because `ChatFunctionCall` struct doesn't have initializer with its properties, it only provides `from:` initializer. So I need to define function call with json format and decode it by `JSONDecoder` if I want to use `ChatFunctionCall` as follows. I think it is unusual and complicated.

```swift
let functionCallJson = """
{
    "name": "add",
    "arguments": "{\\"a\\": \\"1\\", \\"b\\": \\"1\\"}"
}
"""
do {
    if let jsonData = functionCallJson.data(using: .utf8) {
        let chat = Chat(role: .assistant, content: nil, functionCall: try JSONDecoder().decode(ChatFunctionCall.self, from: jsonData))
    }
} catch {
    print(error)
}
```

I can initialize it more simply by implementing initializer with properties for it. So I want to add initializer with properties to `ChatFunctionCall`. I think it is usual flow for initializing struct.

```swift
let functionCall = ChatFunctionCall(name: "add", arguments: "{\"a\": \"1\", \"b\": \"1\"}")
let chat = Chat(role: .assistant, content: nil, functionCall: functionCall)
```

## Affected Areas

`ChatFunctionCall` struct.


Krivoblotsky (approved):
Thanks!

alelordelo:
@ysak-y , did find a way to render the function as “.function” in the chat bubble? Currently it’s wrongly rendered as “.assistant”

Issue #88, opened 2023-07-10
Better error handling for streaming session


KrystofM:
**Is your feature request related to a problem? Please describe.**
Recently I have came over a problem, where my program stopped working getting an error message:
```
keyNotFound(CodingKeys(stringValue: "id", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"id\", intValue: nil) (\"id\").", underlyingError: nil))
```

After some painful debugging I have found out the problem was with my OpenAI key not working.

**Describe the solution you'd like**
Would make sense to implement models for errors as they are described by OpenAI on this page:

https://platform.openai.com/docs/guides/error-codes/api-errors

**Describe alternatives you've considered**
I am not sure why there has not been at least a generic error message printed.

```
var apiError: Error? = nil
            do {
                let decoder = JSONDecoder()
                let object = try decoder.decode(ResultType.self, from: jsonData)
                onReceiveContent?(self, object)
            } catch {
                apiError = error
            }
            
            if let apiError = apiError {
                do {
                    let decoded = try JSONDecoder().decode(APIErrorResponse.self, from: data)
                    onProcessingError?(self, decoded)
                } catch {
                    onProcessingError?(self, apiError)
                }
            }
```

I think the problem concerns this part of the code from StreamingSession.swift . I have not fully investigated this in detail, but my assumption would be that the Error message is not decoded and therefore it returns the error of not being able to decode the ResultType.

When I looked over APIErrorResponse it does seem to be quite generic, so it is indeed interesting that it did not decode itself.

longseespace:
It's my bad. Submitting a new PR, can you check if it fixes your issue @KrystofM : https://github.com/MacPaw/OpenAI/pull/92

KrystofM:
@longseespace Yes seems to be working now. 

Btw kinda funny you responded, since I have been making a chat/embedding PDF type macos app for some time now. Close to release - but probably will just let it for free download possibly open source it, since its just a hobby night time project.

nezhyborets:
Closing as #92 have been merged

PR #87, opened 2023-07-06
Fix issue where chatsStream() throws incorrect error

longseespace:
## What

This PR fixes the regression bug where it throws incorrect error when streaming

Reference: https://github.com/MacPaw/OpenAI/pull/23

## Affected Areas

`OpenAI.chatsStream()`, `StreamingSession`

PR #92, opened 2023-07-25
Fix the issue where it does not decode the Error correctly when streaming response

longseespace:
## What

When using streaming, there might be an error occurring in the middle of the streaming.

This PR fixes it.

Previous PR: https://github.com/MacPaw/OpenAI/pull/87

## Why

Related: https://github.com/MacPaw/OpenAI/issues/88

## Affected Areas

`StreamingSession.swift`

### DIFF
--- a/Sources/OpenAI/Private/StreamingSession.swift
+++ b/Sources/OpenAI/Private/StreamingSession.swift
@@ -74,7 +74,7 @@ final class StreamingSession<ResultType: Codable>: NSObject, Identifiable, URLSe
             
             if let apiError = apiError {
                 do {
-                    let decoded = try JSONDecoder().decode(APIErrorResponse.self, from: data)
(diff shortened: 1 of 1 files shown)


gavtron2000 (approved):

longseespace:
Hey @Krivoblotsky, can you merge this?

Krivoblotsky:
Yup, will add it to the release scope 👌

longseespace:
@Krivoblotsky  Also do you think we should switch to a new implementation. I find the current one quite problematic.

For example, this issue: https://github.com/MacPaw/OpenAI/issues/97

Another issue I encountered is when working with OpenAI-compatible services, they might use SSE comments (started with `:`). It won't work with the current implementation.

Krivoblotsky:
@longseespace let's do it with separate PR to reduce changes size plz 🙏

PR #93, opened 2023-07-30
Custom http headers

benvolioT:
## What

This pull request adds the ability to specify custom HTTP headers when integrating with OpenAI via proxy APIs such as Helicone. It also refactors OpenAi's performRequest() by encapsulating the logic for handling the session data task in its own class, URLSessionDataTaskManager. This should simplify the OpenAI class and make the code easier to manage. This pull request simplifies JSONRequest and MultipartFormDataRequest by abstracting out common code to remove repetition that previously existed in these structs. Finally, it simplifies the protocol URLRequestBuildable, moving configuration-driven parameter setting to the constructors of these struct instances, rather than in the call to that protocol's build() function.

## Why

The motivation for this change was to allow users to integrate with OpenAI via the Helicone proxy. To do so, they not only need to point the OpenAI client at the Helicone host, but they also have to set an additional HTTP header on the requests. This pull request makes that possible.

## Affected Areas

The affected areas of the library include:

* The OpenAI class within the OpenAI.swift file.
* The client configuration within the constructor of the OpenAI class.
* The construction of requests in the following methods: completions(), completionsStream(), images(), embeddings(), chats(), chatsStream(), edits(), model(), models(), moderations(), audioTranscriptions(), and audioTranslations().


benvolioT:
Hi! Is there anything more I need to do in order to get this merged?

Krivoblotsky:
Hey, @benvolioT!
Thanks a lot for the contribution!
Please fix [build](https://github.com/MacPaw/OpenAI/actions/runs/5780727325/job/16100876240?pr=93) and merge conflicts 🙏

Sources/OpenAI/Private/StreamingSession.swift
```diff
@@ -46,40 +48,75 @@ final class StreamingSession<ResultType: Codable>: NSObject, Identifiable, URLSe
...
-        }
-        let jsonObjects = stringContent
+        }                
+        
+        // As data comes in, we process it by trying to decode to JSON. Since we may not have yet received a fully valid
+        // JSON string yet, we keep appending content if we're unable to decode due to a "dataCorrupted" error, trying again
+        // on the next pass in case that new data has completed the string to create valid (parsable) JSON, and so on.
+        // Note that while OpenAI's API doesn't appear to return partial string fragments, proxy service such as Helicone
+        // do seem to do so, making this logic necessary.
+        accumulatedData.append(stringContent)
```
Krivoblotsky, line 58: Could you extract this change to a separate PR, please?
Looks like this is a separate fix, that doesn't apply to custom headers.

Sources/OpenAI/OpenAI.swift
```diff
@@ -50,102 +55,87 @@ final public class OpenAI: OpenAIProtocol {
...
     }
     
     public func audioTranslations(query: AudioTranslationQuery, completion: @escaping (Result<AudioTranslationResult, Error>) -> Void) {
-        performRequest(request: MultipartFormDataRequest<AudioTranslationResult>(body: query, url: buildURL(path: .audioTranslations)), completion: completion)
+        performRequest(request: MultipartFormDataRequest<AudioTranslationQuery, AudioTranslationResult>(body: query, url: buildURL(path: .audioTranslations), headers: generateHeaders(), timeoutInterval: configuration.timeoutInterval), completion: completion)
+    }
+}
+
+extension OpenAI {
+    internal func generateHeaders() -> [String: String] {
```
Krivoblotsky, line 115: I would change the naming to `defaultHeaders` to add the knowledge that it is going to be appended further. 
What do you think?

Krivoblotsky (changes requested):
Overall looks good to me, thanks for the contribution!
1. Please fix the build. 
2. Fix merge conflicts
3. Extract `StreamingSession` fix into a separate PR.
Thanks!

nezhyborets:
Thanks for the contribution @benvolioT ! The lib has drifted away a bit by merging other PRs and working on issues, and `customHeaders` were implemented as part of that work, see https://github.com/MacPaw/OpenAI/pull/263

Issue #101, opened 2023-10-03
Demo App:  Chat Input Allows White Spaces and Enables Send Button Without Text


SarathVSasikala:
**Describe the bug**
The input text in text chats accepts text with white spaces, resulting in **chat bubbles with white spaces**. Additionally, users can click the **send button without entering any text,** causing the creation of empty chat bubbles and generating a response.

**To Reproduce**
1. Go to the demo app chats page
2. Input some text with white spaces
3. Press the send button
4. A chat bubble is created with spaces

**Expected behavior**
After clicking the send button, white spaces should be removed, resulting in chat bubbles without any white spaces.

**Screenshots**
![Simulator Screenshot - iPhone 15 - 2023-10-03 at 14 00 20](https://github.com/MacPaw/OpenAI/assets/146827282/d1bc5410-c615-40f9-a856-ec745eca8dac)

**Additional context**
The chat send button is currently enabled even with an empty chat.

SarathVSasikala:
I recommend adding a validation in text input and send button. 

**Add extra validation in DetailView tapSendMessage function:**
```
let message = inputText.trimmingCharacters(in: .whitespacesAndNewlines)
        if message.isEmpty {
            return
        }
```
also, send button disabledcondition: 
`.disabled(inputText.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty)`

**Output:**


https://github.com/MacPaw/OpenAI/assets/146827282/2c719d74-b92a-4d8b-bdd7-f13f8e0cd203

ingvarus-bc:
Thanks for your rising an issues and your contribution! 🔥
Closing the issue now
Will merge your PR today 🤝

PR #102, opened 2023-10-03
Demo App: Added text chat input validation, enabling 'Send' button only for valid input.

SarathVSasikala:
## What

Implemented chat input text validation in demo app packages. Now, when sending a message, any white spaces in the text input are automatically formatted. Additionally, the send button is disabled when there is no text input to prevent the creation of empty chat bubbles and unwanted responses.

## Why

issue #101 

## Affected Areas

Demo app: Text chats input area.


ingvarus-bc:
Thanks for your contribution! 🔥

ingvarus-bc (approved):

Issue #103, opened 2023-10-10
`"The given data was not valid JSON."` Streaming Error


PallavAg:
**Describe the bug**
I'm on version 0.2.4, while streaming responses I still seem to often be getting the following error:
`"The data couldn't be read because it isn't in the correct format."`

In the console the full output is:
`dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Error Domain=NSCocoaErrorDomain Code=3840 "Unexpected end of file" UserInfo={NSDebugDescription=Unexpected end of file})))`

**To Reproduce**
I usually always get this output when asking GPT-3.5: `"Write 1 to 40 using bullet points"`

Edit: Opened a PR to fix this: #104

LordMatalas:
9 hours ago I had exactly the same situation, noticed that after 30 minutes of inactivity on my app; I was able to get communication with the API without that error message. Now, I'm not experimenting this issue anymore. No changes were made on the app though.

PallavAg:
Ths patch in #97 fixes it! This change would really help the library.

mariusiordache:
I have the same issue. It happens very often. When will the fix be released?

Issue #97, opened 2023-08-16
Partial ChatResult data chunks


jcmourey:
Often receiving partial chunks of ChatResult from an OpenAI streaming session, which causes a DecodingError inside this package.

Example of data received by StreamingSession.urlSession:

data: {"id":"chatcmpl-7oBwbf0NG9vi74Z4mt3XasGhhdAVl","object":"chat.completion.chunk","created":1692196669,"model":"gpt-3.5-turbo-0613","choices":[{"index":0,"delta":{"function_call":{"arguments":" che"}},"finish_reason":null}]}

data: {"id":"chatcmpl-7oBwbf0NG9vi74Z4mt3XasGhhdAVl","object":"chat.completion.chunk","created":1692196669,"model":"gpt-3.5-turbo-0613","choices":[{"index":0,"delta":{"function_call":{"arguments":"et"}},"finish_reason":null}]}

data: {"id":"chatcmpl-7oBwbf0NG9vi74Z4mt3XasGhhdAVl","object":"chat.completion.chunk","created":1692196669,"model":"gpt-3.5-turbo-0613","choices":[{"index":0,"delta":{"function_call":{"arguments":"ah"}},"finish_reason":null}]}

data: {"id":"chatcmpl-7oBwbf0NG9vi74Z4mt3XasGhhdAVl","object":"chat.completion.chunk","created":1692196669,"model":"gpt-3.5-turbo-0613","choices":[{"index":0,"delta":{"function_call":{"arguments":" on"}},"finish_reason":null}]}

data: {"id":"chatcmpl-7oBwbf0NG9vi74Z4mt3XasGhhdAVl","object":"chat.completion.chunk","created":1692196669,"model":"gpt-3.5-turbo-0613","choices":[{"index":0,"delta":{"function_call":{"arguments":" rocket"}},"finish_reason":null}]}

data: {"id":"chatcmpl-

The decoder chokes on the last chunk.

I fixed it in my local copy of the OpenAI package by keeping the incomplete chunk around as leftover for the next round. I don't know enough to know whether it's a valid fix or if there is a better way to address this.

[StreamingSession.patch](https://github.com/MacPaw/OpenAI/files/12360707/StreamingSession.patch)

nezhyborets:
The streaming parser has been completely rewritten, so this issue should be fixed. Closing until reproduced or a similar arises

PR #104, opened 2023-10-13
Correctly handle partial JSON at the end of a chunk

PallavAg:
## What

Streaming a response occasionally encounters the error `"The given data was not valid JSON."`
Related Issues: #97 #103

## Why

The end of a chunk can have an incomplete JSON. 
The rest of the JSON will come through at the beginning of the next chunk.

## Affected Areas

`StreamingSession.swift`

## More Info
This change stores the last JSON from the chunk and appends it to the beginning of the next chunk.

A likely partial JSON is stored when: 
1. `ResultType` decoding fails
2. `APIErrorResponse` decoding fails
3. The JSON is last in the chunk

---

Having an incomplete JSON at the end of a chunk is seemingly random - likely based on OpenAI server load etc.
A request that surfaces this issue more often than not is: `"Write 1 to 40 using bullet points"`

```json
// Sample incomplete chunk
{"id":"chatcmpl-893wTs9tI87Yt0t6moRbyhXT2xn93","object":"chat.completion.chunk","created":1697170797,"model":"gpt-4-0613","choices":[{"index":0,"delta":{"content":"."},"fi
```

---
Kudos to [@jcmourey](https://github.com/jcmourey) for suggesting the fix in #97

### DIFF
--- a/Sources/OpenAI/Private/StreamingSession.swift
+++ b/Sources/OpenAI/Private/StreamingSession.swift
@@ -28,7 +28,7 @@ final class StreamingSession<ResultType: Codable>: NSObject, Identifiable, URLSe
         return session
     }()
     
-    private var prevChunkBuffer = ""
+    private var previousChunkBuffer = ""
 
     init(urlRequest: URLRequest) {
         self.urlRequest = urlRequest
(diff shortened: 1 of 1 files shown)


PSchmiedmayer:
Nice, great addition @PallavAg, we run experience the same issue as noted in #103, this looks like a nice and useful fix, makes sense that it is not guaranteed that each streaming package contains a fully contained JSON.

Would be amazing if this can be merged and added in a next release.

ingvarus-bc:
Neat! 🔥
Amazing addition to the project. What do you think about moving all json handling outside of `urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data)` to a separate method for the sake of testability and more obvious look?

ingvarus-bc:
@PallavAg, hi, I just moved json processing out, what do you think?

ingvarus-bc (approved):

prateekshac1:
It looks like this issue has started to happen again. Have noticed the intensity of this go up in the past 2-3 days. I'm using the latest version of OpenAI 0.2.5 which has his fix. But yet many of my chatsStream(query: query) queries are erroring out with this error 

> "The data couldn’t be read because it isn’t in the correct format. dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Error Domain=NSCocoaErrorDomain Code=3840 "Unable to parse empty data." UserInfo={NSDebugDescription=Unable to parse empty data.})))"

PR #106, opened 2023-10-24
Typo in ReadME

mjdierkes:
max_tokens should reflect the new name maxTokens


ingvarus-bc (approved):
Thanks for the heads up!

Issue #113, opened 2023-11-07
Add support for new multimodal APIs


usagimaru:
Are there any plans to support new APIs for multimodal?

- Text to speech - new API
https://platform.openai.com/docs/api-reference/audio/createSpeech
- Vision - new model and parameters for chat completions
https://platform.openai.com/docs/guides/vision
- DALL-E 3 - new model for image generation
https://platform.openai.com/docs/guides/images

ingvarus-bc:
Hi, thanks for the issue!
> - Vision - new model and parameters for chat completions
> https://platform.openai.com/docs/guides/vision
> - DALL-E 3 - new model for image generation
> https://platform.openai.com/docs/guides/images

Are covered in [PR-115](https://github.com/MacPaw/OpenAI/pull/115)
And you are right, we'll need to add `Create Speech` API support. Thanks!

ingvarus-bc:
@usagimaru, added `Create Speech` feature in [PR-120](https://github.com/MacPaw/OpenAI/pull/120)

PR #120, opened 2023-11-13
Feat: Create Speech  [ tts-1, tts-1-hd ] 

ingvarus-bc:
## What
Implemented [Create Speech](https://platform.openai.com/docs/api-reference/audio/createSpeech) API functionality.

## Why
This new feature was requested in #113 and is a part of latest OpenAI update.

## Affected Areas
Affects root [OpenAI.swift](Sources/OpenAI/OpenAI.swift), [Models.swift](Sources/OpenAI/Public/Models/Models/Models.swift) files.  
Added [AudioSpeechQuery](Sources/OpenAI/Public/Models/AudioSpeechQuery.swift) and [AudioSpeechResult](Sources/OpenAI/Public/Models/AudioSpeechResult.swift)

## More Info
If you got any suggestions to improve this code, feel free to comment ❤️


Sources/OpenAI/Public/Models/AudioSpeechResult.swift
```diff
@@ -0,0 +1,58 @@
...
+        let fileURL = path.appendingPathComponent(filename)
+        try data.write(to: fileURL)
+    }
+    
+    #if canImport(AVFoundation)
+    /// Gets an `AVAudioPlayer` instance configured with the audio data.
+    ///
+    /// - Returns: An `AVAudioPlayer` instance or nil if there is no audio data or if there is issue initializing an `AVAudioPlayer`.
+    /// - Note: Import **AVFoundation**
+    public func getAudioPlayer() -> AVAudioPlayer? {
```
peteliev, line 43: Why do we need to have this method inside `AudioSpeechResult`? As for me, `AVAudioPlayer` is not dedicated to storing audio.

Sources/OpenAI/Public/Models/AudioSpeechResult.swift
```diff
@@ -0,0 +1,58 @@
...
+        guard let data = audioData else {
+            NSLog("No audio data")
+            return nil
+        }
+        
+        do {
+            let audioPlayer = try AVAudioPlayer(data: data)
+            return audioPlayer
+        } catch {
+            NSLog("Error initializing audio player: \(error)")
```
peteliev, line 53: What do you think to use `print()` function instead of `NSLog`. As for me, it's common to use `print()` due to its simplicity and lower performance overhead.

Sources/OpenAI/Public/Models/AudioSpeechResult.swift
```diff
@@ -0,0 +1,58 @@
...
+    public let audioData: Data?
+    
+    /// Saves the audio data to a file at a specified file path.
+    ///
+    /// - Parameters:
+    ///     - name: The name for the file.
+    ///     - format: The format of the audio data, as defined in **`AudioSpeechQuery.AudioSpeechResponseFormat`**.  For example: **`.mp3`**
+    ///     - path: The destination file path as an URL.
+    /// - Throws: Throws an NSError if there is an issue with writing the data to the specified file path.
+    public func saveAs(_ name: String, format: AudioSpeechQuery.AudioSpeechResponseFormat, to path: URL) throws {
```
peteliev, line 25: Why do we need to have `saveAs` method inside `AudioSpeechResult` struct? I need clarification on it.

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -0,0 +1,78 @@
...
+        self.input = input
+        self.voice = voice
+        self.speed = {
+            guard let speed else { return "1.0" }
+            let isSpeedOutOfBounds = speed >= 4.0 && speed <= 0.25
+            guard !isSpeedOutOfBounds else {
+                NSLog("[AudioSpeech] Speed value must be between 0.25 and 4.0. Setting value to closest valid.")
+                return speed < 0.25 ? "1.0" : "4.0"
+            }
+            return String("\(speed)")
```
peteliev, line 74: We can simplify

```swift
return "\(speed)"
```

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -0,0 +1,78 @@
...
+             flac
+    }
+    /// One of the available TTS models: tts-1 or tts-1-hd
+    public let model: Model
+    /// The text to generate audio for. The maximum length is 4096 characters.
+    public let input: String?
+    /// The voice to use when generating the audio. Supported voices are alloy, echo, fable, onyx, nova, and shimmer.
+    public let voice: AudioSpeechVoice
+    /// The format to audio in. Supported formats are mp3, opus, aac, and flac.
+    public let response_format: AudioSpeechResponseFormat
```
peteliev, line 46: Let's use lowerCamelCase naming for properties :)

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -0,0 +1,78 @@
...
+    
+    /// Encapsulates the response formats available for audio data.
+    ///
+    /// **Formats:**
+    /// -  mp3
+    /// -  opus
+    /// -  aac
+    /// -  flac
+    public enum AudioSpeechResponseFormat: String, Codable {
+        case mp3,
```
peteliev, line 34: What do you think use more Swifty syntax?

```swift
public enum AudioSpeechResponseFormat: String, Codable {
   case mp3
   case opus
   case aac
   case flac
}
```

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -0,0 +1,78 @@
...
+
+/// Learn more: [OpenAI Speech – Documentation](https://platform.openai.com/docs/api-reference/audio/createSpeech)
+public struct AudioSpeechQuery: Codable, Equatable {
+    
+    /// Encapsulates the voices available for audio generation.
+    ///
+    /// To get aquinted with each of the voices and listen to the samples visit:
+    /// [OpenAI Text-to-Speech – Voice Options](https://platform.openai.com/docs/guides/text-to-speech/voice-options)
+    public enum AudioSpeechVoice: String, Codable {
+        case alloy,
```
peteliev, line 18: What do you think use more Swifty syntax? 

```swift
public enum AudioSpeechResponseFormat: String, Codable {
   case alloy
   case echo
   case fable
   case onyx
   case nova
   case shimmer
}
```

peteliev (changes requested):
Great addition, but let's polish the code ⭐

Demo/DemoChat/Sources/UI/TextToSpeechView.swift
```diff
@@ -0,0 +1,156 @@
...
+public struct TextToSpeechView: View {
+    
+    @ObservedObject var store: MiscStore
+    
+    @State private var prompt: String = ""
+    @State private var voice: AudioSpeechQuery.AudioSpeechVoice = .alloy
+    @State private var speed: Double = 1
+    @State private var responseFormat: AudioSpeechQuery.AudioSpeechResponseFormat = .mp3
+    
+    private let formats: [AudioSpeechQuery.AudioSpeechResponseFormat] = [.mp3, .aac, .flac, .opus]
```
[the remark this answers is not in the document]
ingvarus-bc: Because all of the properties here a related to the current view and current "ViewModel" is `MiscStore` which handles all the logic of "Misc" tab. Though I agree that it would be better to follow MVVM in a proper way, so it will come with a major update of the demo app, which has to be refactored and improved in a lot of ways.

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -0,0 +1,78 @@
...
+            guard !isModelOfIncorrentFormat else {
+                NSLog("[AudioSpeech] 'AudioSpeechQuery' must have a valid Text-To-Speech model, 'tts-1' or 'tts-1-hd'. Setting model to 'tts-1'.")
+                return .tts_1
+            }
+            return model
+        }()
+        self.input = input
+        self.voice = voice
+        self.speed = {
+            guard let speed else { return "1.0" }
```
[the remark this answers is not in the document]
ingvarus-bc: Agreed 👍

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -0,0 +1,78 @@
...
+            let isSpeedOutOfBounds = speed >= 4.0 && speed <= 0.25
```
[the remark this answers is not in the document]
ingvarus-bc: minSpeed and maxSpeed, that's right

Sources/OpenAI/Public/Models/AudioSpeechResult.swift
ingvarus-bc: Well, to enable devs conveniently be able to save the data to file, as we only receive a raw audio data to file, so unless we use some kind of player that supports raw data, we have to save it as a file. It may be sometimes not obvious what will the request return, so mostly it is file or url to file, but this is the only case where we have raw data, that was my motive to have more flexibility built in.

Sources/OpenAI/Public/Models/AudioSpeechResult.swift
```diff
@@ -0,0 +1,58 @@
+//
+//  AudioSpeechResult.swift
+//
+//
+//  Created by Ihor Makhnyk on 13.11.2023.
+//
+
+import Foundation
+#if canImport(AVFoundation)
```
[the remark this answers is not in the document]
ingvarus-bc: yep, one for import and one for checking if it is available, it is for linux users who don't have all Apple packages (without it linux builds fail)

paramaggarwal:
Is there a way to support streaming of this audio? As documented here; https://platform.openai.com/docs/guides/text-to-speech/streaming-real-time-audio

ingvarus-bc:
> Is there a way to support streaming of this audio? As documented here: https://platform.openai.com/docs/guides/text-to-speech/streaming-real-time-audio

Hi, yes, it seems like it is possible, because the data is actually being streamed every time we make a request and if you do like this:
```
curl https://api.openai.com/v1/audio/speech \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "tts-1",
    "input": "The quick brown fox jumped over the lazy dog.",
    "voice": "alloy"
  }' \
  --output speech.mp3
```
You can actually see progress how is file being fetched and written. I think for that we need to have some audio player that supports partial raw data playback, to stream the result right into it.

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
peteliev: Great, thanks!

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -299,8 +301,10 @@
 				CODE_SIGN_ENTITLEMENTS = App/Demo.entitlements;
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
+				DEVELOPMENT_TEAM = YX43KN96E7;
```
peteliev, line 304: Do we need a predefined Development Team?

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -43,36 +43,59 @@ public struct AudioSpeechQuery: Codable, Equatable {
...
         
         self.model = {
-            guard let model else { return .tts_1 }
-            let isModelOfIncorrentFormat = model != .tts_1 && model != .tts_1_hd
-            guard !isModelOfIncorrentFormat else {
-                NSLog("[AudioSpeech] 'AudioSpeechQuery' must have a valid Text-To-Speech model, 'tts-1' or 'tts-1-hd'. Setting model to 'tts-1'.")
-                return .tts_1
-            }
-            return model
+            AudioSpeechQuery.validateSpeechModel(model)
```
peteliev, line 65: What is the main idea of validating model it inside closure? 

```swift
self.model = AudioSpeechQuery.validateSpeechModel(model)
```

ingvarus-bc: Oh, good point 🔥
Thanks)

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -43,36 +43,59 @@ public struct AudioSpeechQuery: Codable, Equatable {
...
+        guard !isModelOfIncorrentFormat else {
+            NSLog("[AudioSpeech] 'AudioSpeechQuery' must have a valid Text-To-Speech model, 'tts-1' or 'tts-1-hd'. Setting model to 'tts-1'.")
+            return .tts_1
+        }
+        return inputModel
     }
+    
+    private static func validateSpeechSpeed(_ inputSpeed: Double?) -> String {
+        guard let inputSpeed else { return "1.0" }
+        let minSpeed = 0.25
```
peteliev, line 91: What do you think about moving it to to a higher level?

```swift
struct AudioSpeechQuery {
   ...
   
   private enum Constants {
      static let minSpechSpeed = 0.25
      ...
   }
   
   ...
}
```

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -43,36 +43,59 @@ public struct AudioSpeechQuery: Codable, Equatable {
...
+        let maxSpeed = 4.0
+        let isSpeedOutOfBounds = inputSpeed >= maxSpeed && inputSpeed <= minSpeed
+        guard !isSpeedOutOfBounds else {
+            NSLog("[AudioSpeech] Speed value must be between 0.25 and 4.0. Setting value to closest valid.")
+            return inputSpeed < minSpeed ? "1.0" : "4.0"
```
peteliev, line 96: And then use `Constants`

Sources/OpenAI/Public/Models/AudioSpeechQuery.swift
```diff
@@ -43,36 +43,59 @@ public struct AudioSpeechQuery: Codable, Equatable {
...
+        guard let inputModel else { return .tts_1 }
+        let isModelOfIncorrentFormat = inputModel != .tts_1 && inputModel != .tts_1_hd
+        guard !isModelOfIncorrentFormat else {
+            NSLog("[AudioSpeech] 'AudioSpeechQuery' must have a valid Text-To-Speech model, 'tts-1' or 'tts-1-hd'. Setting model to 'tts-1'.")
+            return .tts_1
+        }
+        return inputModel
     }
+    
+    private static func validateSpeechSpeed(_ inputSpeed: Double?) -> String {
```
peteliev, line 89: Let's rename this method. As for me, validation is about the bool result. I suggest you to use `normalizeSpeechSpeed`. It looks much more apparent.

Sources/OpenAI/Public/Models/AudioSpeechResult.swift
```diff
@@ -25,7 +25,7 @@ public struct AudioSpeechResult {
     public func saveAs(_ name: String, format: AudioSpeechQuery.AudioSpeechResponseFormat, to path: URL) throws {
         guard let data = audioData else {
             throw NSError(
```
peteliev, line 27: What do you think about using Swift Error instead of Objective-C NSError. Do we have some limitations?

Demo/DemoChat/Sources/SpeechStore.swift
```diff
@@ -0,0 +1,60 @@
...
+                                          audioPlayer: player,
+                                          originResponse: response,
+                                          format: query.responseFormat.rawValue)
+            audioObjects.append(audioObject)
+        } catch {
+            print(error.localizedDescription)
+        }
+    }
+    
+    func getFileInDocumentsDirectory(_ data: Data, fileName: String, _ dir: @escaping (URL) -> Void) {
```
peteliev, line 48: Let's rename the parameter from `dir` to `completion`, `completionBlock`, or something else.

Demo/DemoChat/Sources/UI/TextToSpeechView.swift
```diff
@@ -0,0 +1,155 @@
...
+                                    player.volume = 1
+                                    player.play()
+                                }
+                            }, label: {
+                                Image(systemName: "play.fill").foregroundStyle(object.format == AudioSpeechQuery.AudioSpeechResponseFormat.opus.rawValue ? Color.secondary : Color.accentColor)
+                            })
+                        }
+                        .swipeActions(edge: .trailing, allowsFullSwipe: false) {
+                            Button {
+                                presentUserDirectoryDocumentPicker(for: object.originResponse.audioData!, filename: "GeneratedAudio.\(object.format)")
```
peteliev, line 120: Let's don't use force unwrapping :)

Demo/DemoChat/Sources/UI/TextToSpeechView.swift
```diff
@@ -0,0 +1,155 @@
...
+        store.getFileInDocumentsDirectory(audioData, fileName: filename) { fileUrl in
+            let filePickerVC = UIDocumentPickerViewController(forExporting: [fileUrl], asCopy: false)
+            filePickerVC.shouldShowFileExtensions = true
+            
+            guard let vc = getCurrentViewController() else { return }
+            vc.present(filePickerVC, animated: true, completion: nil)
+        }
+    }
+    
+    private func getCurrentViewController() -> UIViewController? {
```
peteliev, line 149: What do you think about moving it as extension for `View`? As for me, it is great solution not only for `TextToSpeechView`

peteliev (approved):
Brilliant job 💎

PR #122, opened 2023-11-14
New content format

rawnly:
## What
Includes changes from #115

`content` of `ChatQuery` is now `ChatContent`.

[Link to the api spec](https://platform.openai.com/docs/api-reference/chat)

## Missing 
Some tests. `tools` / `tool_calls` / `content_filter`. 

## Problems
One thing I noticed is that even tho the available format for the content is `Array<ChatContent>/String` the response received from the api will always be a string. 

I tested this branch on a real application and sometimes I have decoding errors, even tho analyzing the response I really can't tell what's wrong.

## Affected Areas
Chat related models


rawnly:
I'd love to hear some feedback about the `ChatContent`. Swift is not my main language so I might have done something dumb or over-engineered 😅

ingvarus-bc:
Addressing the issue about decoding errors, can't it be the case with partial json receiving, the issue which was just recently solved [here](https://github.com/MacPaw/OpenAI/pull/104)?

Additionally, what do you think about this OpenAI [Chat Chunk Object](https://platform.openai.com/docs/api-reference/chat/streaming)? As it says in docs, it seems to be returning chat completion now, that is why it is just a string.
chat_return

I may be wrong, but it seems intuitive to me that if we are about to receive mixed content, it should be done by streaming methods instead of completions. 🤔

rawnly:
> Addressing the issue about decoding errors, can't it be the case with partial json receiving, the issue which was just recently solved [here](https://github.com/MacPaw/OpenAI/pull/104)?

Absolutely I missed that! I'll take another spin to check if this happens again

Sources/OpenAI/Public/Utilities/CodableUtilities.swift
```diff
@@ -0,0 +1,118 @@
...
+///         struct FullName {
+///             firstName: String
+///             lastName: String
+///         }
+///     }
+///
+///     let person = Person(name: .string("John Doe"))
+///     let fullNamePerson = Person(name: .object(.init(firstName: "John", lastName: "Doe")))
+///     ```
+public enum StringOrCodable<T: Codable>: Equatable, Codable where T: Equatable {
```
rawnly, line 41: [Looking at the api](https://platform.openai.com/docs/api-reference/chat/create) looks like only the role `user` can actually send arrays.


CleanShot 2023-11-14 at 12 13 57@2x
> ChatContent with `user` role

CleanShot 2023-11-14 at 12 13 33@2x
> ChatContent with `assistant` role

This should be fine with the actual implementation, I tried to handle both via `StringOrCodable`

rawnly: i'm not sure if there's a cleaner way to handle this

Tests/OpenAITests/OpenAITestsDecoder.swift
```diff
@@ -235,12 +271,16 @@ class OpenAITestsDecoder: XCTestCase {
...
             choices: [
-                .init(index: 0, message:
-                        Chat(role: .assistant,
-                             functionCall: ChatFunctionCall(name: "get_current_weather", arguments: nil)),
-                      finishReason: "function_call")
+                .init(
+                    index: 0,
+                    // TODO: Fix "Ambiguous use of 'init(role:content:name:functionCall:)'" when omitting content
+                    // maybe by removing `init` the overload?
+                    message: Chat(role: .assistant, content: nil as String?, functionCall: .init(name: "get_current_weather", arguments: nil)),
```
rawnly, line 310: I tried to keep existing constructors for backwards compatibility but this ended up like this
`content: nil as String?` . If you try to omit the content it returns error because is too generic

ingvarus-bc: I got the issue, it was in ChatQuery having a lot of inits. What do you think about reducing it to:
```swift
   public init(role: Role, content: Codable? = nil, name: String? = nil, functionCall: ChatFunctionCall? = nil) {
        let stringOrCodable: StringOrCodable<[ChatContent]>?
        
        if let string = content as? String {
            stringOrCodable = .string(string)
        } else if let arr = content as? [ChatContent] {
            stringOrCodable = .object(arr)
        } else {
            stringOrCodable = nil
        }
        
        self.init(role: role, contents: stringOrCodable, name: name, functionCall: functionCall)
   }
```
And with this kind of constructor, the error you mentioned here will be solved and you'll be able to omit `content` like this:
```swift
Chat(role: .assistant, functionCall: .init(name: "get_current_weather", arguments: nil)
```

Sources/OpenAI/Public/Models/ChatQuery.swift
```diff
@@ -7,10 +7,104 @@
 
 import Foundation
 
+
+
+// See more https://platform.openai.com/docs/guides/text-generation/json-mode
+public struct ResponseFormat: Codable, Equatable {
```
ingvarus-bc, line 13: Would be nice to bring it out to the separate file. As I understand it relates only to chats, so let's perhaps name it smth like `ChatQueryResponseFormat` or `ChatQueryJSONResponse` 🤔

ingvarus-bc:
@rawnly I must really highlight your amazing work here, thank you, for your contribution and that you care about the project! ✨ 

There is a lot going on in code, will take me some time to get into it, what can tell as of now, we definitely need to break down everything that is going on inside [ChatQuery.swift](Sources/OpenAI/Public/Models/ChatQuery.swift) into separate files as OpenAI have a lot of things encapsulated in one another in their Chat API. And it is best to preserve everything that is already there as it is, just "deprecate" it upon release. 

*ps. Do you still have decoding errors after [the fix](https://github.com/MacPaw/OpenAI/pull/104)?*

rawnly:
> ps. Do you still have decoding errors after [the fix](https://github.com/MacPaw/OpenAI/pull/104)?

looks like it's fixed now 🚀

Tests/OpenAITests/OpenAITestsDecoder.swift
rawnly: awesome i'll try this asap

rawnly: fixed in  https://github.com/MacPaw/OpenAI/pull/122/commits/791484c93590f48e11d65f42c80a3258bc3ea9ea

Sources/OpenAI/Public/Models/ChatQuery.swift
rawnly: what about `ChatQuery.ResponseFormat`?

ingvarus-bc: Alright, but let's make it as an extension for more clarity and code being more swiftly 🙌

rawnly: I'd split the chat part into separate files too, maybe into a subfolder smth like 
CleanShot 2023-11-14 at 13 56 38@2x

rawnly:
Btw one thing i noticed is that new models kind of ignore the `functions` passed to them, maybe it's worth support both for backwards compatibility and mark functions as deprecated(?)

Sources/OpenAI/Public/Models/ChatQuery.swift
ingvarus-bc: Perfect! ❤️

ingvarus-bc:
> Btw one thing i noticed is that new models kind of ignore the `functions` passed to them, maybe it's worth support both for backwards compatibility and mark functions as deprecated(?)

Yes, totally agree, that what I was suggesting, here:
> There is a lot going on in code, will take me some time to get into it, what can tell as of now, we definitely need to break down everything that is going on inside [ChatQuery.swift](https://github.com/MacPaw/OpenAI/pull/Sources/OpenAI/Public/Models/ChatQuery.swift) into separate files as OpenAI have a lot of things encapsulated in one another in their Chat API. And it is best to preserve everything that is already there as it is, just "deprecate" it upon release.

rawnly:
Now it should be more clear where things are and it should be easier to manage new models too

rawnly:
Regarding tools etc, what do you think about smth like 

```swift
    Tool(
        type: .function,
        value: .function(
            Tool.Function(
                name: "get_weather",
                description: "Get the weather" 
            )
        )
    )
```

`parameters` is still `JSONObject` but now i added an `.empty` static let to allow no params (which is the default value in `Tool.Function` constructor

ingvarus-bc:
> Regarding tools etc, what do you think about smth like
> 
> ```swift
>     Tool(
>         type: .function,
>         value: .function(
>             Tool.Function(
>                 name: "get_weather",
>                 description: "Get the weather" 
>             )
>         )
>     )
> ```
> 
> `parameters` is still `JSONObject` but now i added an `.empty` static let to allow no params (which is the default value in `Tool.Function` constructor

Is this the one from message/tool_calls/ ?
image

rawnly:
nop it's the one from here https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

ingvarus-bc:
> nop it's the one from here https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

Got you, yeah, cool. There is one additional field for parameters which function accepts 🤔
I am wondering if we can reuse the same `tools` in messages as well

rawnly:
> I am wondering if we can reuse the same `tools` in messages as well

let me push this code so you can look at the model code :)

Sources/OpenAI/Public/Models/Chat/Chat.swift
```diff
@@ -0,0 +1,72 @@
+//
+//  File.swift
+//  
+//
+//  Created by Federico Vitale on 14/11/23.
+//
+
+import Foundation
+
+public struct Chat: Codable, Equatable {
```
ingvarus-bc, line 10: I guess it should be called `Message` as it is being used in messages array

Sources/OpenAI/Public/Models/Chat/Chat.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+    
+    enum CodingKeys: String, CodingKey {
+        case role
+        case content
+        case name
+        case functionCall = "function_call"
+    }
+    
+    public init(role: Role, content codable: Codable? = nil ,name: String? = nil, functionCall: ChatFunctionCall? = nil) {
```
ingvarus-bc, line 33: What do you think about typealiasing this `Codable?` to something like `StringOrChatContent`(as a suggestion, maybe you'll come up with something better), because I am testing it now and it is now very clear what to pass there 🤔

rawnly: absolutely nice idea!

rawnly: maybe also throwing an error if passing the wrong codable?

ingvarus-bc: yeah, that would be neat 🔥

ingvarus-bc, line 33:
```suggestion
    public init(role: Role, contents codable: Codable? = nil ,name: String? = nil, functionCall: ChatFunctionCall? = nil) {
```
Let's make it as well content**S** to have matching constructors and not confuse future users ☺️

ingvarus-bc:
From what I can see, it is indeed only possible to pass functions with assistant role. Did not quite fully understand how these functions supposed to work yet, I got working but rather weird result 😄:
image

ingvarus-bc:
> ### Problems
>
> One thing I noticed is that even tho the available format for the content is Array<ChatContent>/String the response received from the api will always be a string.
>
> I tested this branch on a real application and sometimes I have decoding errors, even tho analyzing the response I really can't tell what's wrong.

Hi, @rawnly! I've been testing different cases and everything seems to be working fine, except that we cannot receive an array as a response, as you mentioned previously, I tried making requests with response_format: json_object, but it did not seem to work, it produces a decoding error. 🤔 I think it is worth looking in that direction.
Screenshot 2023-11-15 at 15 17 09

ingvarus-bc:
What do you think of adding one more level of abstraction for chats method, it is the most complex and the most customisable one and there are a lot of cases when incorrect setting will cause an error for sure. Plus there is "stream" API boolean property, which we can implement on method level and select this way what response to receive a ready message response or streamDeltas for ChatStreamResult.

rawnly:
> What do you think of adding one more level of abstraction for chats method, it is the most complex and the most customisable one and there are a lot of cases when incorrect setting will cause an error for sure

what are you thinking about?

ingvarus-bc:
> what are you thinking about?

I have two things in mind: 
1. A couple of protocol level methods which will build the Chat query according to the settings and properties provided, to be mindful about the errors which we can predict for sure.
2. Or we can add some error handling and fallbacks inside of ChatQuery constructor.

Why I think about it, is because OpenAI documentation for chats is clearly not very straightforward and doesn't cover most of the cases, so devs who will be just figuring it out may have a lot of struggle with these settings and properties.

What do you think?

ingvarus-bc:
@rawnly, hey, we are on our way to make a release tag today, do you have some work in progress or is it okay if I push a couple of things here? ✨

rawnly:
> What do you think ?

makes sense

> A couple of protocol level methods which will build the Chat query according to the settings and properties provided, to be mindful about the errors which we can predict for sure

you mean something like a builder?

ingvarus-bc:
> you mean something like a builder?

yep, sort of 👍
I am gonna push slight refactoring and will deprecate completions (as they're 'legacy' now and chats fully cover their functionality) and will merge the PR

And as of the builder, I think it will be a good addition in the next patch

may I be missing something, what do you think, have you encountered any issues in scope of this PR?🤔

rawnly:
> > you mean something like a builder?
> 
> yep, sort of 👍 I am gonna push slight refactoring and will deprecate completions (as they're 'legacy' now and chats fully cover their functionality) and will merge the PR
> 
> And as of the builder, I think it will be a good addition in the next patch
> 
> may I be missing something? what do you think, have you encountered any issues in scope of this PR?🤔

lgtm can you test the tools in a real request? I didn't have the time and now i'm not at home 👀

rawnly:
Current problem is 
https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices

bcherry:
I'm using this PR in a project and I found that if you pass a single instance of ChatContent, rather than array of [ChatContent], it will just set the content to nil and fail only at request time with a cryptic "content and role must not be nil" message.  The ergonomics of this seem problematic to me, since the swift type  checker is unable to help. Maybe at least allowing a single instance of ChatContent to be passed, and then converted to an array, would be good. Otherwise maybe best to not offer an initializer that accepts `StringOrChatContent` (aka any Codable), so the user isn't exposed to the API variability under the hood. They would either use the explicit string-based or explicit array-of-content-based initializer.

Sources/OpenAI/Public/Models/Chat/Message.swift
```diff
@@ -0,0 +1,129 @@
...
+        self.functionCall = nil
+        self.toolCalls = []
+    }
+    
+    public init(role: Role, content: StringOrCodable<[ChatContent]>? = nil, name: String? = nil, functionCall: ChatFunctionCall?) {
+        self.role = role
+        self.content = content
+        self.name = name
+        self.functionCall = functionCall
+        self.toolCalls = []
```
bcherry, line 91: this initializer is a trap for vision requests - you'll get an invalid request error from sending an empty array in `toolCalls` (which isn't supported for this model yet). You need to use the other initializer with explicit `toolCalls: nil`.

Would it be safe to default to nil here?

Sources/OpenAI/Public/Models/Chat/Message.swift
```diff
@@ -0,0 +1,129 @@
+//
+//  Message.swift
+//  
+//
+//  Created by Federico Vitale on 14/11/23.
+//
+
+import Foundation
+
+public struct Message: Codable, Equatable {
```
bcherry, line 10: This is a pretty generic name for a type - I found it to be a problem in my project that already had its own `Message` type defined, and `OpenAI.Message` didn't work because the `OpenAI` module also contains a class called `OpenAI` ([see this discussion on this potential flaw in Swift](https://forums.swift.org/t/fixing-modules-that-contain-a-type-with-the-same-name/3025/25))

The only thing that worked for me was renaming the type in my own project so `Message` was available for this package to use.

ingvarus-bc: Good point, I think it would be a good idea to use some naming pattern, to avoid confusions, though this would mean that all projects using this package would need to update namings.

Sources/OpenAI/Public/Models/Chat/ChatTool.swift
```diff
@@ -0,0 +1,191 @@
...
+                        debugDescription: "Invalid data encountered when decoding StringOrCodable<T>"
+                    )
+                )
+            }
+        }
+    }
+}
+
+public struct ToolCall: Codable, Equatable {
+    public let index: Int
```
sroyaura, line 105: I'm getting decoding errors because `index` is not present in the response.
```
▿ DecodingError
  ▿ keyNotFound : 2 elements
    - .0 : CodingKeys(stringValue: "index", intValue: nil)
    ▿ .1 : Context
      ▿ codingPath : 5 elements
        - 0 : CodingKeys(stringValue: "choices", intValue: nil)
        ▿ 1 : _JSONKey(stringValue: "Index 0", intValue: 0)
          ▿ rep : Rep
            - index : 0
        - 2 : CodingKeys(stringValue: "message", intValue: nil)
        - 3 : CodingKeys(stringValue: "tool_calls", intValue: nil)
        ▿ 4 : _JSONKey(stringValue: "Index 0", intValue: 0)
          ▿ rep : Rep
            - index : 0
      - debugDescription : "No value associated with key CodingKeys(stringValue: \"index\", intValue: nil) (\"index\")."
      - underlyingError : nil
```

oxy9en:
Hi Guys. Do you have any updates about this PR?

kalafus:
main is in sync with api specification as of [pull](https://github.com/MacPaw/OpenAI/pull/169) (excepting that Edits endpoint is obsolete, betas not implemented).

PR #142, opened 2024-01-02
Update OpenAI.swift

0xHexE:


0xHexE:
Closing since there is already similar PR open

PR #143, opened 2024-01-04
Create Contributions

DanteAusonio:
## What

5 key additions to the project:
1. Conversation Model allowing full conversation history to be stored in a single object. Includes methods for appending system, assistant, and user messages. Includes methods for chat history containing only user and ai messages, and chat history containing a history buffer (last n messages that fall with in token limit)
2. String Extension to allow for token counting on any string.
3. Memory Models. Use Chat GPT to generate different types of memory. Can be used dynamically in Agents thanks to Strategy Design Pattern.
4. Chat System & Phase Models. A chat system is an LLM system that allows the conversational assistant to change its behavior as the conversation progresses. This is accomplished using an Injector Agent (see #5). Each system has a set of phases which represent the set of roles the conversational ai can take on. 
5. Injector Agent: This observational ai performs check ins on a user's conversation with the conversational ai. It determines which of the phases from the chat system the conversational ai should enter, and injects system prompts to dynamically change the direction of the conversation and behavior of the conversational agent as the conversation progresses. The injector Agent is a conversational system of my own creation that is very versatile and can result in far more complex conversations, but also serves as an example use of the memory models, as well as Chat System and Phase models.

## Why
I made these contributions to the project because I think it would be powerful to start implementing langchain type functionality. We have the basics down, now we can start adding higher levels of abstraction. The Conversation Model and String token count Extension seemed like necessary additions. The conversation model is a straight forward and robust way to store full conversations, and allows for easy addition of new messages, as well as is well suited for UI implementation. 

## Affected Areas

Nothing preexisting was changed. I only added new files and objects on top of what was already there


kalafus:
Adding functionality to the OpenAI API is a great idea, but what about using MacPaw/OpenAI as a dependency? This would prevent scope creep of this project as well as guarantee decoupling from MacPaw/OpenAI implementation details.

nezhyborets:
Hi @DanteAusonio . Thanks for your contribution! Please correct me as I'm not an LLM pro, but it seems to introduce notions that are related to, but not a part of, OpenAI Platform's docs and guides. Integrating such changes may require more work and thought than integrating usual OpenAI changes. I believe we would also have to clearly outline the functionality in README, and show its usage in demo. 

Having said that, it seems that this functionality may be really helpful to users. An approach could be to break it down and integrate piece by piece. You know the code in PR better than I do. Do you think we could extract something that is "core" and most useful, strip away anything else for now? Or it only goes as an interconnected bundle? If a single useful piece can be extracted, what would that be? I'm sure we can come up with an integration approach, your replies and input would be really helpful. Thanks!

DanteAusonio:
Hey, ya someone already got back to me on this a while ago..this kind of
functionality would likely be good on its own as a different package that
uses this one as a dependency.

nezhyborets:
Hi @DanteAusonio , thanks for the quick reply! I see, still, thanks for this PR, I learned new things today 😅, maybe someday we'll put some of the functionality into life in one form or another. Going to close this PR

PR #149, opened 2024-01-27
trimming contentString

frogcjn:
stringContent may contain a prefix new line ("\n") before the first "data:" separator.

## What

When query with chat streaming API, the response may contain a prefix new line (`\n`) before the first `data:` separator.

## Affected Areas

Chat Streaming APIs

### DIFF
--- a/Sources/OpenAI/Private/StreamingSession.swift
+++ b/Sources/OpenAI/Private/StreamingSession.swift
@@ -57,7 +57,7 @@ final class StreamingSession<ResultType: Codable>: NSObject, Identifiable, URLSe
 extension StreamingSession {
     
     private func processJSON(from stringContent: String) {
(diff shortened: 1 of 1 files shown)


ingvarus-bc (approved):

ingvarus-bc:
Hi, @frogcjn, thanks for your interest in the project and for your PR.
It seems like character trimming for each data component is already being done on line 63:
```swift
.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
```
So, I am not exactly sure we need to do that twice. What do you think?

frogcjn:
There are some cases there is "\n" before the "data："， and nothing before "\n", then it will raise error since the first element after the second trimming is empty.

ingvarus-bc:
Got you, thanks!

Issue #128, opened 2023-11-21
The data couldn’t be read because it isn’t in the correct format.


TalentDev0401:
I am implementing OpenAI into my iOS app.
I integrated OpenAI library into my iOS app using swiftpackage.
I configured OpenAI api key and implemented "Completion" with prompt text.
But, I have got above issue.

Below are my code.

let openAI = OpenAI(apiToken: "YOUR_TOKEN_HERE")
let query = CompletionsQuery(model: .textDavinci_003, prompt: "What is 42?", temperature: 0, maxTokens: 100, topP: 1, frequencyPenalty: 0, presencePenalty: 0, stop: ["\\n"])
openAI.completions(query: query) { result in
 switch result {
    case .success(let result):
        print(result.choices)
    case .failure(let error):
        //Handle chunk error here
       print(error.localizedDescription)
    }
}

But, I have got error message - "The data couldn’t be read because it isn’t in the correct format."
How can I solve this issue shortly?

JanusChoi:
Have the same issue today while using ChatStream

svkozak:
Started suddnly getting this error too (when streaming), although it was supposedly fixed in #104

neelvirdy:
Also started getting this error suddenly today when streaming

prateekshac1:
I've started to suddenly get this error while using chatStream too, and very often. A bit concerning as my app is in production. Trying to find a quick fix.  Any suggestions?

svkozak:
This looks like the best fix: https://github.com/MacPaw/OpenAI/pull/151#issuecomment-1913703132

And the reason is that if `stringContent` starts or ends with a separator (in our case `data:`), the `.components(separatedBy:)` adds an empty string to the array ([reference](https://developer.apple.com/documentation/swift/stringprotocol/components%28separatedby:%29-8gl9t#discussion))

prateekshac1:
Perfect. Yes, that fixes it, and makes sense. Thank you @svkozak :)

iamklim:
Same here, receive this error while using `chatsStream`

nezhyborets:
Closing because it seems to be fixed with the PRs mentioned above, e.g. #151 , #154

PR #151, opened 2024-01-28
fix: fix json decode error when jsonContent is emtpy

oblank:
## What

ignore empty string when json decode, it may case some issure, such as [https://github.com/MacPaw/OpenAI/issues/128](https://github.com/MacPaw/OpenAI/issues/128)

## Why

fix bug

## Affected Areas

json decode when openai response

### DIFF
--- a/Sources/OpenAI/Private/StreamingSession.swift
+++ b/Sources/OpenAI/Private/StreamingSession.swift
@@ -57,6 +57,9 @@ final class StreamingSession<ResultType: Codable>: NSObject, Identifiable, URLSe
 extension StreamingSession {
     
     private func processJSON(from stringContent: String) {
+        if stringContent.isEmpty {
+            return
+        }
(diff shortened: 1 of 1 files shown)


ActuallyTaylor:
I also found this issue, however I think a much better fix is to just change: 
```
let jsonObjects = "\(previousChunkBuffer)\(stringContent)"
    .components(separatedBy: "data:")
    .filter { $0.isEmpty == false }
    .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
```

to:

```
let jsonObjects = "\(previousChunkBuffer)\(stringContent)"
    .components(separatedBy: "data:")
    .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
    .filter { $0.isEmpty == false }
```

Simply moving the filter down by one line fixes this and does not require any other changes or checks.

ingvarus-bc (approved):

ingvarus-bc:
Thanks for your contribution! 🔥

PR #156, opened 2024-02-05
array thread safety wrapper to fix racing condition causing Fatal error

kalafus:
thread safe array wrapper to fix racing condition

## What

`Fatal error: UnsafeMutablePointer.moveUpdate(from:) with negative count` at `self?.streamingSessions.removeAll(where: { $0 == object })`

## Why

array can be concurrently accessed with append and removeAll methods.

## Affected Areas

networking


ingvarus-bc (approved):

ingvarus-bc:
Nice job! Thanks for your PR!

PR #157, opened 2024-02-05
StreamableQuery parity

kalafus:
make StreamableQuery symmetrically convert ChatQuery objects to non-streamable mode on the fly

## What

ChatQuery initialized with `stream = true` can be used in `.chats`

## Why

parity with non-streamable StreamableQuery being made streamable as needed.

## Affected Areas

ChatQuery


kalafus:
as it stands, the presumption is that ChatQuery will be created as non-streamable, and needs be made streamable on-demand. could also, for example, take away the `stream` parameter at ChatQuery initialization. that might make more sense, as the stream parameter is perfunctory if it is being changed as needed.

another option could be to eliminate both the default to `stream: false` as well as the whole concept of the `Streamable` protocol. at present, setting `stream` to `false` is an illusion -- there is no such thing in this API, seeing as a new struct is created with `stream: true` whenever needed.

 thoughts?

nezhyborets:
Hi @kalafus , thanks for the contribution! By looking at existing `StreamableQuery` and `chats`-`chatsStream` methods of `OpenAI`, I'd say it's kind of strange that `var stream: Bool` on `ChatStream` is public, but at the same we internally set it to what we need inside `chats` and `chatsStream`, basically ignoring whatever was set by user.

I think in general it would be better to change it, so that user controls stream-ability by either setting value on `Query` or by calling a method that would do that for them.

But I'm also not sure if we want to do this change now. It works fine and there is other stuff that needs attention at the moment. So I'd say we merge it to make the existing solution more complete, and think about refactoring later

nezhyborets (approved):

PR #160, opened 2024-02-05
remove legacy completions endpoint

kalafus:
## What

remove legacy completions endpoint (uses obsolete models and truncates responses)

## Why

legacy `.completions(query: CompletionsQuery)` endpoint technically still works, but it frequently truncates responses. the `.chats(query: ChatQuery)` endpoint is available and not janky.

## Affected Areas

`.completions(query: CompletionsQuery)`


kalafus:
Would it be better to pull for a deprecation message?
`@available(*, deprecated, message: "endpoint obsolete; please consider using the chat/completions endpoint.")`

nezhyborets (approved):
👍👍👍

PR #161, opened 2024-02-05
view abstraction of navigationTitle and modelSelect for reuse in demo

kalafus:
## What

view abstraction of navigationTitle and modelSelect for reuse in demo

## Why

feature parity and consistent appearance

## Affected Areas

TTS, Chat (via DetailView)


ingvarus-bc (approved):
🔥🔥🔥

PR #162, opened 2024-02-05
update moderations results keywords, abstract display of moderations keywords

kalafus:
## What

update moderations results keywords, abstract display of moderations keywords

## Why

4 moderations keywords were missing

## Affected Areas

ModerationsResults, demo


ingvarus-bc:
Hi, @kalafus, thanks for your PR! 🔥
Looks great! Though, what do you think of keeping everything in CamelCase? 
Just for consistency, so that the code is in more Swifty way

kalafus:
sure thing.

got a more concise way to do the Iterator? Reflection is no longer a concise way to flit through these.

Sources/OpenAI/Public/Models/ModerationsResult.swift
```diff
@@ -26,25 +35,93 @@ public struct ModerationsResult: Codable, Equatable {
...
                 case sexualMinors = "sexual/minors"
                 case violence
                 case violenceGraphic = "violence/graphic"
             }
+
+            public func makeIterator() -> CategoriesIterator {
+                return CategoriesIterator(self)
+            }
+
+            public struct CategoriesIterator: IteratorProtocol {
```
ingvarus-bc, line 57: Hi, addressing your question:
> got a more concise way to do the Iterator? Reflection is no longer a concise way to flit through these.
  
What do you think of perhaps using Mirror? Maybe something like this:
```swift
public struct CategoriesIterator: IteratorProtocol {
     public typealias Element = (String, Bool)

      private let categories: Categories
      private var properties: [(String, Bool)]

      public init(_ categories: Categories) {
           self.categories = categories
           self.properties = Mirror(reflecting: categories).children.compactMap {
                 guard let name = $0.label, let value = $0.value as? Bool else { return nil }
                 return (name, value)
           }
       }

       public mutating func next() -> Element? {
            guard !properties.isEmpty else { return nil }
            return properties.removeFirst()
       }
}
```

kalafus: @ingvarus-bc , good thought. The trouble with doing it that way is, as far as I can determine, it can only reflect the camelCase names, and not the CodingKeys translations. Hence the long form... Unless you can figure out a way of looking up the CodingKeys from the camelCase names, which I couldn't.

kalafus: OK, got it with reflection. Relying on the order of CodingKeys being same as the order of variables in the structs. Checking this with DEBUG asserts for future verification.

ingvarus-bc (approved):
Nice job there 🔥
Thank you!

PR #164, opened 2024-02-06
follow JSON API

kalafus:
## What

JSON API uses snake_case

## Why

enhanced clarity when CodingKeys can be removed

## Affected Areas

Query & Result structs


kalafus:
Code Duplication is pre-existing; all of it is in Tests.

SunburstEnzo:
Sorry but I disagree, Swift uses camel case for property names, it should stay consistent in the event a different encoding is used

kalafus:
OK, thanks for explaining!

PR #168, opened 2024-02-07
Codable > Encodable | Decodable

kalafus:
## What

be specific with Codable:
use
Encodable for queries,
Decodable for results,
instead of Codable for everything

## Why

accurate inheritance prevents accidental misuse

## Affected Areas

JSON encoding and decoding during networking

## NOTE

I don't know what to do with the .stub() functions, which require that Responses are Encodable for testing purposes, when Responses should only have to be Decodable.


ingvarus-bc:
Hey, @kalafus, thanks for your contribution! ✨
You're absolutely right about:
> accurate inheritance prevents accidental misuse
   
   
It seems like tests here require some fixing now too 🙏🏼

kalafus:
got an error that extensions for the Encodable protocol won't automatically implement this way... but it's done.

nezhyborets:
I'm sorry that I have to discard so much work. But this doesn't look like it would provide that much value. Add limitations to tests. Make things "unusual" (usually stuff is just Codable). And also takes away code generation. But maybe I don't see the full picture. What misuse cases do you have in mind? Closing for now but we can get back to it if we decide so

PR #172, opened 2024-02-14
streamline `performSpeechRequest`

kalafus:
## What

remove special handling of `performSpeechRequest`

handle receiving raw (non-JSON-wrapped) data in the same workflow as all other incoming data.

## Why

`AudioSpeechResult` wrapper uniquely does not represent JSON wrapped Data, but the endpoint rather returns raw Data.

Since it's not JSON encoded, there's no way to use JSONDecoder and init(Decoder:) to correctly handle the returned data, but exceptional handling can be added for Data return type to skip JSONDecoder and return the raw Data object. Exception first attempts to decode using builtin Data type JSONDecoder, and only returns raw data when Data type is specified, and JSONDecode as Data type has failed, so this is a pretty reusable approach.

## Affected Areas

performSpeechRequest


nezhyborets:
Hi @kalafus . Thanks for the contribution! Sorry, but I don't agree with the new solution. Instead of a focused separate piece of code that could be tailored to endpoint needs, we put everything into one piece of code and want to handle all the cases. Consider scale - what if there are hundreds of endpoints with all the different kinds of response formats? We can discuss what bothers you with the current solution and see if we can find an approach. If it's code duplication - we can try to look into composition. If it's naming - we could rename it, but actually for new it seems like speech response format IS a special case. When we have more endpoints with exactly the same response formats we could generalize, but until then it'd be a guess shot

PR #177, opened 2024-02-20
:alien: (Compatability) Make model optional to allow using TogetherAI endpoint

djmango:
## What

TogetherAI's API mocks OpenAI's, however the model is not returned in chat stream responses, thus it must be made optional in order to proceed without erroring.

Simply added a ? flag to make model optional.

Enables you to use the TogetherAI API like so:
```swift
let configuration = OpenAI.Configuration(token: "NOTYOURTOKEN", host: "api.together.xyz", timeoutInterval: 10)
let ai = OpenAI(configuration: configuration)
```

## Affected Areas

ChatStreamResponse.swift


kalafus:
Does this fall under this scope-creep? This project specifically implements the OpenAI API.

djmango:
It's an extremely minor change, and most of the inference providers intentionally clone the OpenAI API in order to be compatible with project such as this. I have submitted a ticket to them regarding this issue as well, but this solution will be more robust.

Sources/OpenAI/Public/Models/ChatStreamResult.swift
```diff
@@ -124,7 +121,7 @@ public struct ChatStreamResult: Codable, Equatable {
     /// Each chunk has the same timestamp.
     public let created: TimeInterval
     /// The model to generate the completion.
-    public let model: String
+    public let model: String?
```
kalafus, line 124: this is the only change relevant to the intent of the pull

djmango:
I've now added support for paths in the model host config, enabling Groq API support (api.groq.com/openai)
```swift
extension OpenAI {
    func buildURL(path: String) -> URL {
        var components = URLComponents()
        components.scheme = "https"

        // Check if host contains a path component
        let hostParts = configuration.host.split(separator: "/", maxSplits: 1, omittingEmptySubsequences: true)
        if hostParts.count > 1 {
            // Host contains a path component
            components.host = String(hostParts[0])
            let hostPath = "/" + hostParts[1]
            components.path = hostPath + path
        } else {
            // Host does not contain a path component
            components.host = configuration.host
            components.path = path
        }

        guard let url = components.url else {
            fatalError("Unable to construct URL from components.")
        }

        return url
    }
}
```

DanielhCarranza:
Any update on this I'm interested in this because many models can be use this way, Grow, TogetherAI, Huggingface, Azure, etc

djmango:
It works yeah, I use it with Groq, TogetherAI, all the LiteLLM supported models (Anthropic mostly). My fork will be updated to maintain that as well as OpenAI, unsure if they will merge this into master here.

djmango:
branching far away from the original scope of this project now

PR #183, opened 2024-03-19
PCM support for text-to-speech

vilcsak:
## What

I'm not sure if it's a recent change but the API supports `pcm` for text-to-speech and it isn't supported by this library. A simple change to allow `pcm` as a format.

## Why

Bring up to spec with the API

## Affected Areas

Text-to-speech

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,7 +548,7 @@ public struct AudioSpeechQuery: Codable, Equatable {
 **Response:**
 
 ```swift
-/// Audio data for one of the following formats :`mp3`, `opus`, `aac`, `flac`
+/// Audio data for one of the following formats :`mp3`, `opus`, `aac`, `flac`, `pcm`
 public let audioData: Data?
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -548,7 +548,7 @@ public struct AudioSpeechQuery: Codable, Equatable {
 **Response:**
 
 ```swift
-/// Audio data for one of the following formats :`mp3`, `opus`, `aac`, `flac`
+/// Audio data for one of the following formats :`mp3`, `opus`, `aac`, `flac`, `pcm`
```
SunburstEnzo, line 551: Interestingly the documentation also mentions `wav`, so I'm wondering if we should add that as well to this PR or a separate PR. 

image

SunburstEnzo (approved):
Thanks for the PR! Just a comment on whether it's worth also adding `wav` support while we're here.

nezhyborets (approved):
Great, thanks! And yes, if OpenAI supports wav I see no reason why we shouldn't cut that functionality. I'll merge this so we move forward, but I created a task for wav https://github.com/MacPaw/OpenAI/issues/242

README.md
nezhyborets: Answered in the approval comment

PR #190, opened 2024-04-02
Add pathPrefix to Configuration

batanus:
## What
Add `pathPrefix` to OpenAI.Configuration.

## Why
This is needed for reverse proxy servers - for example, if the proxy server has the next path: `some-proxy-server.com/api`, and it will be passed as `host` - `func buildURL(path: String) -> URL` method will crash because `some-proxy-server.com/api` is not a valid host. 

The fix for this issue is to add `pathPrefix`, which will be added before each `path` when building a URL. So in provided example, `some-proxy-server.com/api` will be splitter to `host: "some-proxy-server.com", pathPrefix: "/api"`

## Affected Areas
Added new property `pathPrefix` to `OpenAI.Configuration`

### DIFF
--- a/Sources/OpenAI/OpenAI.swift
+++ b/Sources/OpenAI/OpenAI.swift
@@ -23,16 +23,18 @@ final public class OpenAI: OpenAIProtocol {
         /// API host. Set this property if you use some kind of proxy or your own server. Default is api.openai.com
         public let host: String
         public let port: Int
+        public let pathPrefix: String
         public let scheme: String
         /// Default request timeout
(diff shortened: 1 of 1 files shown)


kkmoskalenko:
Great addition, had to make the same changes to my own fork to support Groq endpoints like:
https://api.groq.com/openai/v1/chat/completions

Hope your PR gets approved soon 🙏

011235813 (approved):

kkmoskalenko (approved):

nezhyborets:
This option has been implemented using `basePath`

PR #191, opened 2024-04-02
Add SSLDelegateProtocol to StreamingSession

batanus:
## What
Add `SSLDelegateProtocol` delegate to `StreamingSession`

## Why
I've discovered that when initializing `OpenAI` class with `URLSession`, it isn't utilized if the request uses `StreamingSession` (which is logical). Consequently, if a client wishes to implement `SSL Pinning` in their application and passes their `URLSession` with a set delegate (which implements the `didReceive challenge: URLAuthenticationChallenge` method from `URLSessionDelegate`), `URLSessionDelegate` methods will be invoked only for requests that **don't** use `StreamingSession` and the streaming mechanism.

This PR introduces a mechanism that allows the user to pass their `delegate` inherited from `SSLDelegateProtocol` and manually handle `didReceive challenge: URLAuthenticationChallenge`.

## Affected Areas
`OpenAI` and `StreamingSession` classes initializers


nezhyborets (approved):
This is great, thanks for the contribution!

PR #193, opened 2024-04-06
Made Token Requirement optional

DominatorVbN:
## What

In favour of supporting LLAMA CPP Server, where token is not needed made a small change that makes authentication optional

## Why

As LLAMA CPP server when hosted in local does not require an token to be present in header, this changes were made.

## Affected Areas

Authentication


nezhyborets:
Implemented in https://github.com/MacPaw/OpenAI/pull/280

PR #214, opened 2024-06-03
 feat: implement cancellable

tisfeng:
## What

Enable user to cancel stream request, I just picked up the commit from  https://github.com/longseespace/OpenAI/commit/a290a1a119fc465cdb1740472242fbfca57069b9#diff-2a16b8014d00e2223611951c839756f88dd24a58d6d2d34962fa9c329eab6580

## Why

Currently stream requests cannot be canceled once they are sent, which is bad.


nezhyborets:
Implemented in https://github.com/MacPaw/OpenAI/pull/251

Issue #221, opened 2024-08-01
base_url support instead of host, port, and scheme and move '/v1' to base_url


daisukes:
When I want to connect to an OpenAI compatible proxy server (i.e. litellm), it does not accept requests including `/v1` in the URL.

- This works `http://localhost:8000/chat/completion`
- This does not work `http://localhost:8000/v1/chat/completion`

The official OpenAI python library has `base_url` option to create client and the default `base_url` contains `/v1`

- https://github.com/openai/openai-python/blob/195c05a64d39c87b2dfdf1eca2d339597f1fce03/src/openai/_client.py#L121
```python
        if base_url is None:
            base_url = f"https://api.openai.com/v1"
```

I can make a client for my proxy like

```python
client = openai.OpenAI(api_key="api-key", base_url="http://localhost:8000")
```

So, I want `base_url` support instead of host, port and scheme configuration here.

https://github.com/MacPaw/OpenAI/blob/843e087929aa806adb611dbca93f9a4a7f28be04/Sources/OpenAI/OpenAI.swift#L30

Also, I want APIPath not including `/v1` and move it to `base_url`

https://github.com/MacPaw/OpenAI/blob/843e087929aa806adb611dbca93f9a4a7f28be04/Sources/OpenAI/OpenAI.swift#L211-L231

codeAndxv:
I need this feature. Can a manager review this PR and merge it into the code?

nezhyborets:
Implemented in https://github.com/MacPaw/OpenAI/pull/268

PR #222, opened 2024-08-01
feat: Add configurable endpoint option

udostm:
## What

This PR adds fully customizable endpoint support. Instead of "just" adding `basePath` you can now pass a fully configured endpoint URL.

## Why

Internal/hosted OpenAI Instances sometimes use different paths to host the OpenAI endpoint. Previous versions of the framework only allowed for changing `scheme`, `host` and `port`. As changing the `basePath` is a common requirement (and also supported by frameworks in other languages) this PR adds fully customisable endpoint support - incl. the `basePath`.

## Affected Areas

- Introduced a new initializer that accepts a fully configured `endpoint: URL`.
- The existing initializer (accepting `host`, `port`, and `scheme`) remains available and now also configures the endpoint.
- Converted the previously stored `host`, `port`, and `scheme` properties to computed properties based on the endpoint to maintain API compatibility.

### DIFF
--- a/Sources/OpenAI/OpenAI.swift
+++ b/Sources/OpenAI/OpenAI.swift
@@ -37,6 +37,7 @@ final public class OpenAI: OpenAIProtocol {
             components.scheme = scheme
             components.host = host
             components.port = port
+            components.path = "/v1"
 
             self.endpoint = components.url!
             self.timeoutInterval = timeoutInterval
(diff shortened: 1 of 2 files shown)


udostm:
Base on #221 it might make sense to rewrite the code to not include the `/v1` suffix implicitly?

nezhyborets:
Thanks for the contribution, sorry but it got outdated due to other changes. All this functionality seems to have been implemented

PR #243, opened 2025-01-21
fix: filter out sse comments

tisfeng:
## What

Filter out comments, like ": OPENROUTER PROCESSING"

## Why

FIX OpenRouter error: https://github.com/tisfeng/Easydict/issues/743

### DIFF
--- a/Sources/OpenAI/Private/StreamingSession.swift
+++ b/Sources/OpenAI/Private/StreamingSession.swift
@@ -60,14 +60,23 @@ extension StreamingSession {
         if stringContent.isEmpty {
             return
         }
-        let jsonObjects = "\(previousChunkBuffer)\(stringContent)"
+
+        // Join the previous chunk buffer with the new chunk, and filter out comments.
+        // FIX OpenRouter: https://github.com/tisfeng/Easydict/issues/743
+        let filteredContent = "\(previousChunkBuffer)\(stringContent)"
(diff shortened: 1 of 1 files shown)


Sources/OpenAI/Private/StreamingSession.swift
```diff
@@ -60,14 +60,23 @@ extension StreamingSession {
...
             return
         }
-        let jsonObjects = "\(previousChunkBuffer)\(stringContent)"
+
+        // Join the previous chunk buffer with the new chunk, and filter out comments.
+        // FIX OpenRouter: https://github.com/tisfeng/Easydict/issues/743
+        let filteredContent = "\(previousChunkBuffer)\(stringContent)"
             .trimmingCharacters(in: .whitespacesAndNewlines)
+            .components(separatedBy: .newlines)
+            .filter { !$0.hasPrefix(":") }  // Filter out comments, like ": OPENROUTER PROCESSING"
```
Krivoblotsky, line 69: Hey, @tisfeng!
Is there any documentation on OpenRouter API about this line? We'd like to be sure we don't get a regression here and do not break any other providers.

Krivoblotsky: And what do you think about writing unit tests that cover both cases (with and without comments)

nezhyborets:
Implemented in https://github.com/MacPaw/OpenAI/pull/259

PR #245, opened 2025-01-24
Feat/o1

nezhyborets:
## What
Add o1 and o1-mini to Model list

## Why
To enable support for these models in SDK

## Affected Areas
`Model.swift`


Krivoblotsky (approved):
Thanks!

PR #247, opened 2025-01-26
Feat: Add developer role

dounan:
## What

Add support for the `developer` Chat completion message role.
https://cdn.openai.com/spec/model-spec-2024-05-08.html#definitions

## Why

Align with latest model spec

## Affected Areas

Chat Completion API

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatQuery.swift
+++ b/Sources/OpenAI/Public/Models/ChatQuery.swift
@@ -111,6 +111,7 @@ public struct ChatQuery: Equatable, Codable, Streamable {
     public enum ChatCompletionMessageParam: Codable, Equatable {
 
         case system(Self.SystemMessageParam)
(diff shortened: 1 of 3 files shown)


Sources/OpenAI/Public/Models/ChatQuery.swift
```diff
@@ -119,6 +120,8 @@ public struct ChatQuery: Equatable, Codable, Streamable {
             switch self {
             case .system(let systemMessage):
                 return Self.UserMessageParam.Content.string(systemMessage.content)
+            case .developer(let developerMessage):
+                return Self.UserMessageParam.Content.string(developerMessage.content)
```
nezhyborets, line 124: Looks strange that we map unrelated content to `UserMessageParam.Content`, but I see that it has been implemented this way for other cases, so this is a question for another day

nezhyborets (approved):
Great, thanks!

PR #248, opened 2025-01-26
Bug: Add initializer for ToolCallParam.FunctionCall

dounan:
## What

Add an initializer for ToolCallParam.FunctionCall

## Why

Cannot initialize without it. I have a use case that requires me to initialize a ToolCallParam (instead of using one returned by the SDK).

## Affected Areas

Chat Completion API

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatQuery.swift
+++ b/Sources/OpenAI/Public/Models/ChatQuery.swift
@@ -539,6 +539,14 @@ public struct ChatQuery: Equatable, Codable, Streamable {
                     public let arguments: String
                     /// The name of the function to call.
                     public let name: String
+
+                    public init(
+                        arguments: String,
+                        name: String
+                    ) {
+                        self.arguments = arguments
+                        self.name = name
+                    }
                 }
             }
         }


dounan:
My app talks to various LLM backends so I have an internal representation of the chat that I then map to the request format of the LLM API I need.

In order to map to the OpenAI SDK request format, I have code that looks like this:

```
func adapt(_ toolCall: AIResponse.ToolCall) -> ChatQuery.ChatCompletionMessageParam.AssistantMessageParam.ToolCallParam {
        switch toolCall {
        case .function(let fnToolCall):
            return .init(
                id: fnToolCall.id,
                function: .init(
                    arguments: fnToolCall.argumentsJSON,
                    name: fnToolCall.name
                )
            )
        }
    }
```

Without this PR, I cannot initialize the value of the `function` field . I am open to alternatives though if there is a better way.

nezhyborets:
Thanks for the explanation! Let's go with this for now so that you can do what you need. With time we may gather more knowledge about related requests and maybe come up with another solution (if needed)

nezhyborets (approved):

PR #252, opened 2025-02-01
[deprecated] Support comments in streaming response

SplittyDev:
## What

This PR adjusts the streaming response parsing logic to gracefully deal with comments.

## Why

Server Sent Events, the standard used by the OpenAI API and compatible APIs, support comments by definition[^1], but the parsing code did not deal with them gracefully. Instead, it caused a decoding error, trying to parse the comment as JSON.

Even though the official OpenAI API does not seem to be using comments as far as I can tell, other OpenAI-compatible APIs are using them (e.g. OpenRouter), and they are part of the standard, so should probably be supported anyway.

Honestly, that part of the code probably needs a bigger overhaul in the future, because the parsing logic there is far from robust, and far from correct. It's correct enough to handle the specific way that OpenAI is using Server Sent Events, but it ignores many parts of the spec, such as multi-line events that can have multiple `data:` parts while still being the same event.

The changes in this PR are just a very small part of what needs to be done in order to fully and correctly support Server Sent Events, but it's enough to resolve the issue I was facing with OpenRouter, so I'll leave it at that for now.

[^1]: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream

## Affected Areas

Only the stream response parsing is affected.

### DIFF
--- a/Sources/OpenAI/Private/StreamInterpreter.swift
+++ b/Sources/OpenAI/Private/StreamInterpreter.swift
@@ -1,0 +1,78 @@
+//
+//  StreamInterpreter.swift
+//  OpenAI
+//
+//  Created by Oleksii Nezhyborets on 03.02.2025.
+//
+
+import Foundation
+
+/// https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation
+/// 9.2.6 Interpreting an event stream
+class StreamInterpreter<ResultType: Codable> {
+    private let streamingCompletionMarker = "[DONE]"
+    private var previousChunkBuffer = ""
+    
+    var onEventDispatched: ((ResultType) -> Void)?
+    
+    func processData(_ data: Data) throws {
+        guard let stringContent = String(data: data, encoding: .utf8) else {
+            throw StreamingError.unknownContent
+        }
+        try processJSON(from: stringContent)
+    }
+    
+    private func processJSON(from stringContent: String) throws {
+        if stringContent.isEmpty {
+            return
+        }
+
+        let fullChunk = "\(previousChunkBuffer)\(stringContent)"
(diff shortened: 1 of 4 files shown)


nezhyborets:
Great! Thanks for the contribution! Do you think any unit tests may be of value over this piece of code?

SplittyDev:
> Great! Thanks for the contribution @spittydev! Do you think any unit tests may be of value over this piece of code?

It would probably be better to have some tests for it, yes.

I didn't add any tests myself because there is no baseline to test against. It doesn't seem like streaming responses are being tested at all right now, so it didn't seem to be a priority, and in all honesty I was just trying to fix the issue at hand.

I can look into adding some tests for streaming responses, but I'm not sure how fast I can get it done. Might be a few days until I find the time.

Let me know what you think!

SplittyDev:
@nezhyborets I finally had the time to add a test for streaming responses.

nezhyborets:
Hi @SplittyDev . Thanks. Have you seen my PR to your repo? https://github.com/FiveSheepCo/OpenAI-OpenRouter/pull/1
As I wasn't able to push directly, did it through PR

SplittyDev:
@nezhyborets Ahh sorry, I did not see that! If you don't mind, I'll close this PR and open a new one from another branch in a few minutes

nezhyborets:
@SplittyDev sure

nezhyborets:
@SplittyDev let's somehow combine both your and mine recent changes. From mine I'd like to keep that Interpreter is extracted. So maybe use my StreamInterpreter file and put your changes in it. As for tests - I don't mind if you don't use my tests if you don't see added value in them over your tests

SplittyDev:
@nezhyborets let's continue in #259 please 👍

PR #266, opened 2025-02-17
Bug: Fix error parsing in StreamInterpreter

nezhyborets:
## What

Try parsing APIError on the whole response instead of later on divided pieces

## Why

To fix [the issue](https://github.com/MacPaw/OpenAI/issues/261#issuecomment-2663022284)

## Affected Areas

`StreamInterpreter`. Also, while debugging, I added an ability to demo to specify **non-streaming** way of completing conversations. Didn't add a UI to control that though. Added a task for that https://github.com/MacPaw/OpenAI/issues/265

PR #267, opened 2025-02-18
Fix StreamInterpreter error handling for API error responses in streaming mode

wangqi:
## What

This PR improves error handling in StreamInterpreter.swift when processing streaming responses. Previously, if the remote server returned an API error, the JSON parsing would fail, and a parsing error would be thrown instead of the actual API error.

Now, the implementation explicitly checks for an "error" key in the response JSON. If detected, it parses the content as an API error and returns the real error message instead of a generic parsing failure.

## Why

- When working in streaming mode, the API may return an error response that does not conform to the expected message format.
- Without this fix, the client receives a JSON parsing error instead of the actual API error, making debugging difficult.
- This ensures that API errors are correctly detected and propagated, improving error handling and developer experience.

## Affected Areas

- Sources/OpenAI/Private/StreamInterpreter.swift
- Affects how errors are handled when receiving streaming responses from the OpenAI API.

### DIFF
--- a/Sources/OpenAI/Private/StreamInterpreter.swift
+++ b/Sources/OpenAI/Private/StreamInterpreter.swift
@@ -16,11 +16,6 @@ class StreamInterpreter<ResultType: Codable> {
     var onEventDispatched: ((ResultType) -> Void)?
     
     func processData(_ data: Data) throws {
-        let decoder = JSONDecoder()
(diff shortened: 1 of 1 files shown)


nezhyborets:
Hi @wangqi , thanks for the contribution! Sorry that we kinda missed each other, the fix is already release https://github.com/MacPaw/OpenAI/pull/266. Let me know if you'd like something to be appended to the solution in #266

PR #268, opened 2025-02-19
feat: Add custom api version support

LimChihi:
## What

This PR introduces a new `apiVersion` property in the `OpenAI.Configuration` structure . This allows users to customize the version segment (e.g. v123, replacing the default v1) in the API endpoint URLs. For example, requests can now be directed to https://api.openai.com/v2/... or other versions by setting apiVersion: "v2" in the configuration.

## Why

- Flexibility: Some users may need to interact with non-standard or future versions of the API (e.g., self-hosted OpenAI-compatible services that use a different versioning scheme).
- Future-Proofing: If OpenAI deprecates or updates the API version path, clients can seamlessly adapt without waiting for library updates.
- Enterprise Use Cases: Organizations running custom deployments of OpenAI-like APIs often use versioned endpoints beyond v1, and this change simplifies integration.

## Affected Areas

- Configuration: Added the apiVersion property to `OpenAI.Configuration`, including default value handling.
- Network Requests: Updated URL construction logic to use the configured apiVersion instead of hardcoding v1 in endpoint paths.
- Tests: `OpenAITests.testCustomApiVersionBuilt` unit tests to cover the new apiVersion behavior.

### DIFF
--- a/Sources/OpenAI/OpenAI.swift
+++ b/Sources/OpenAI/OpenAI.swift
@@ -25,6 +25,9 @@ final public class OpenAI {
 
         /// Optional base path if you set up OpenAI API proxy on a custom path on your own host. Default is ""
         public let basePath: String
+        
+        /// API version if you want to use a specific version of the API. Default is "v1"
+        public let apiVersion: String
 
         public let port: Int
         public let scheme: String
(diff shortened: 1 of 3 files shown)


Krivoblotsky (approved):
Nice, thanks for the contribution!

nezhyborets:
Hi @LimChihi , thanks for the contribution! This functionality should definitely be added. I'm a bit worried that if we fix the order of components in this way: `configuration.basePath, configuration.apiVersion`, it may be not flexible enough for developers need. Thus, I was thinking, that maybe we should just use existing `basePath` for versioning. Like, by default `basePath` would be `/v1`

```swift
final public class OpenAI {
    public struct Configuration {       
        public init(
            token: String,
            organizationIdentifier: String? = nil,
            host: String = "api.openai.com",
            port: Int = 443,
            scheme: String = "https",
            basePath: String = "/v1",
            timeoutInterval: TimeInterval = 60.0
        ) { ... }
        ...
    }
}
```

if a developer wants to customise `basePath` it would also be their responsibility to correctly specify version depending on their needs. We can refer to [Swagger spec](https://swagger.io/docs/specification/v2_0/api-host-and-base-path/) in docs. What do you think?

LimChihi:
@nezhyborets You're absolutely righ, by adding default value `/v1` to `basePath` may be the best way at beginning.  

But now would risk breaking existing integrations where developers have already customized `basePath` (They'll lose the `/v1` path).

**Proposed Solution:**

- Add the  `apiVersion` property (defaulting to `v1`) and keep `basePath` as-is. It will lose some flexibility.
- Chaning the `basePath` default value to `v1`. This is best for overall, but can lead to API breaks.

Which option do you think would be more reasonable? Happy to refine further!

nezhyborets:
BasePath has been introduced to this SDK recently, maybe if we make the change now, it would not break that many references, hopefully not that many people have adopted it yet. I think we better go through this inconvenience sooner than later

nezhyborets:
@LimChihi BasePath has been introduced to this SDK recently, maybe if we make the change now, it would not break that many references, hopefully not that many people have adopted it yet. I think we better go through this inconvenience sooner than later

LimChihi:
@nezhyborets Hi I changed the `basePath` default value to '/v1', Please review the pr! (^_^)

nezhyborets (approved):

nezhyborets:
@LimChihi great, thanks! I think we should release it right away

PR #269, opened 2025-02-22
feat:add stream_options to query

LimChihi:
#### What  

This PR introduces **`ChatQuery.streamOptions`** and its **`includeUsage`** property, enabling developers to optionally retrieve `usage` metadata (e.g., token consumption) **after streaming responses complete**. The feature is disabled by default and requires explicit opt-in via `includeUsage: true`.

#### Why  

https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156

https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options

#### Affected Areas  

1. **Core Logic**  
   - `ChatQuery`: Added `streamOptions` configuration and associated type definitions.  
   - `ChatStreamResult`: Added `usage` property.
2. **Testing**  
   - Added unit tests verifying:  
     - `testChatQueryWithStreamOptions`
     - `testChatQueryWithoutStreamOptions`
3. **Backward Compatibility**  
   - Zero breaking changes – existing users must **opt-in** to enable `includeUsage`.

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatQuery.swift
+++ b/Sources/OpenAI/Public/Models/ChatQuery.swift
@@ -74,6 +74,8 @@ public struct ChatQuery: Equatable, Codable, Streamable {
     /// If set, partial message deltas will be sent, like in ChatGPT. Tokens will be sent as data-only server-sent events as they become available, with the stream terminated by a data: [DONE] message.
     /// https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format
     public var stream: Bool
+    /// Options for streaming response. Only set this when you set stream: true.
+    public var streamOptions: Self.StreamOptions?
 
     public init(
         messages: [Self.ChatCompletionMessageParam],
@@ -95,7 +97,8 @@ public struct ChatQuery: Equatable, Codable, Streamable {
         topLogprobs: Int? = nil,
         topP: Double? = nil,
         user: String? = nil,
-        stream: Bool = false
+        stream: Bool = false,
+        streamOptions: StreamOptions? = nil
     ) {
         self.messages = messages
         self.model = model
@@ -117,6 +120,7 @@ public struct ChatQuery: Equatable, Codable, Streamable {
         self.topP = topP
         self.user = user
         self.stream = stream
+        self.streamOptions = streamOptions
     }
 
     public enum ChatCompletionMessageParam: Codable, Equatable {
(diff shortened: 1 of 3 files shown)


nezhyborets (approved):
So now there's usage in both ChatStream and it's streaming counterpart. Great, thanks!

nezhyborets:
@LimChihi looks like the build failed, could you check?

LimChihi:
@nezhyborets I'll check later. I'm going to lunch now.😂😂

LimChihi:
@nezhyborets Hi, all checks had passed. I also added the `includeUsage` parameter to `chatsStream(query:)`, that will make calling more convenience.

nezhyborets (changes requested):
Oh, it changes the shape of a core method... This kinda makes the API of the SDK lean into certain way. I know for some other similar methods we have params, like `assistantId`, `before`, `after`. But that's because those params are passed as part of url, not as part of query (body). 

Let's keep it the way it is for now and only add `includeUsage` to Query interface. But I'll keep in mind that such a change may add convenience. And maybe we'll reshape main interface so that query values are more controllable using params on methods instead of values of `Query`

Or is this change required by the current state of things, by how `Streamable` works in our SDK?

nezhyborets:
@LimChihi I see, it's unavoidable with current state of things, as `stream: true` is controlled not by what a developer sets on `ChatQuery`, but by a method they use.

But let's make it more generic. Let's pass the whole `StreamOptions` as a parameter to `chatStream`. This way it would kinda reflect OpenAI API: if we use `stream` we can specify `stream_options`

nezhyborets:
My the way, just to discuss. We have a problem that `ChatQuery`

nezhyborets:
After this PR is merged I'd probably create another one to append the doc, something like this:
```swift
public struct ChatQuery: Equatable, Codable, Streamable {
    ...
    // This value is overridden when using `chat` and `chatStream` methods of SDK.
    public var stream: Bool

    ...

    // This value is overridden when using `chat` and `chatStream` methods of SDK.
    public var streamOptions: StreamOptions
}
```

Looks ugly, but it is what it is at the moment. And I don't even know how to change it. Only subclassing comes to mind, like:
```swift
class ChatStreamQuery: ChatQuery {
    let streamOptions: StreamOptions
}

protocol OpenAIAsync {
    func chatsStream(query: ChatStreamQuery) -> AsyncThrowingStream<ChatStreamResult, Error>
}
```

LimChihi:
@nezhyborets hi, I revert the commit. We can support the feature first, Then we can discuss better and more convenient solutions

LimChihi:
In current api, You need to set the stream parameters in two places.
```swift
// 1
let query = ChatQuery(streamOptions: .init(includeUsage: true))
// 2
openAI.chatsStream(query: query)
```
maybe we can changed to 

```swift
// non stream
let query = ChatQuery()
openAI.chat(query: query)

// stream 
let query = ChatStreamQuery()
openAI.chat(query: query)
```
it will config parameters in one places. And two Query confirm to on Protocol.

nezhyborets:
Thanks again for the cooperation!

PR #282, opened 2025-03-10
Make `created: TimeInterval` optional

JackuXL:
## What

Make `created: TimeInterval` in `ModelResult.swift` optional

## Why

/models of Deepseek api does not return `created` field
```json
{
  "object": "list",
  "data": [
    {
      "id": "deepseek-chat",
      "object": "model",
      "owned_by": "deepseek"
    },
    {
      "id": "deepseek-reasoner",
      "object": "model",
      "owned_by": "deepseek"
    }
  ]
}
```

## Affected Areas
OpenAI.models
ModelResult


nezhyborets:
Hi @JackuXL , thanks for the PR! I am not sure we can change `ModelResult` structure in this way though, as it would conflict with the OpenAI's API Reference, where `created` is declared as non-optional. But I understand a problem and there should be a solution. I am thinking about providing some method where a custom structures could be used as Query and Result. Something like this:
```swift
func performRequest<ResultType: Codable>(query: any Codable, completion: @escaping (Result<ResultType, Error>) -> Void) -> CancellableRequest
```
But make it public.
This way you could declare your own `DeepSeekModelsQuery` and `DeepSeekModelResult` in any way you like and call it on the SDK. This way the main subset of types (like `ChatResult`, `ModelResult` etc.) would be 100% compatible with OpenAI API Reference, and any number of custom Query/Result types could be added for any other provider (on a developer time for the time being, maybe somewhen we'd add such types into SDK)

SplittyDev:
@nezhyborets Ahh I just noticed that I opened basically the same PR under #286. Google AI Studio also doesn't support the `created` field.

For my particular use-case, a custom structure wouldn't cut it either, because my app supports both custom endpoints and openRouter. So I basically have to go with a "catch-all" approach and just try to cover as many bases as possible.

PR #286, opened 2025-03-16
Make `ModelResult/created` optional

SplittyDev:
## What

This PR makes the `created` field on `ModelResult` optional.

## Why

Google AI Studio doesn't support the `created` field on `Model`, which is why listing models fails.

## Affected Areas

Only the `created` field of `ModelResult` is modified. This is technically a breaking change.


SplittyDev:
Ah sorry, this seems to be a duplicate of #282. Closing.

Issue #261, opened 2025-02-05
Streaming does not honor http status codes


SplittyDev:
**Describe the bug**
When streaming responses (e.g. `client.chatsStream(query:)`), the stream does _not_ throw when encountering an HTTP error response. That is not only counterintuitive, but also catastrophically wrong.

It breaks the expectation that errors on the server end will cause the stream to throw, makes errors _undetectable_ and client-side error handling _impossible_, as any `for let result in await client.chatsStream(...)` loops just silently break, indicating that everything went well.

**To Reproduce**
Stream a response from a server that returns an http error, for example `500`.

Here's a short snippet that exhibits this issue:
```swift
do {
  let query: ChatQuery = someValidChatQuery // define a valid `ChatQuery` here
  for try await response in client.chatsStream(query: query) {
    print("Received response")
  }
  print("Finished without error")
} catch {
  print("Error")
}
```

In the case that the server responds with an HTTP error, the "Error" message will never be printed. Instead, you will just see the "Finished without error" message.

**Expected behavior**
The `AsyncThrowingStream` should throw.

**Additional context**

I have a possible fix, but I'm not sure I like it very much:
```diff
diff --git a/Sources/OpenAI/Private/StreamingSession.swift b/Sources/OpenAI/Private/StreamingSession.swift
index fa763f5..379f892 100644
--- a/Sources/OpenAI/Private/StreamingSession.swift
+++ b/Sources/OpenAI/Private/StreamingSession.swift
@@ -46,6 +46,15 @@ final class StreamingSession<ResultType: Codable>: NSObject, Identifiable, URLSe
             onProcessingError?(self, error)
         }
     }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) {
+        if let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode >= 400 {
+            self.onProcessingError?(self, StreamingError.statusError(httpResponse.statusCode))
+            completionHandler(.cancel)
+            return
+        }
+        completionHandler(.allow)
+    }
     
     private func subscribeToParser() {
```

I've extended `StreamingError` too:
```diff
diff --git a/Sources/OpenAI/Private/StreamingError.swift b/Sources/OpenAI/Private/StreamingError.swift
index 66f77ed..5043fa5 100644
--- a/Sources/OpenAI/Private/StreamingError.swift
[...]

PR #287, opened 2025-03-16
Handle HTTP status errors in streaming sessions

SplittyDev:
## What

This PR adds proper HTTP error handling to `StreamingSession`, and exposes access to both the `HTTPURLResponse` as well as the `statusCode` via `OpenAIError/statusError`.

## Why

HTTP status errors (404, 500, etc.) aren't currently handled in streaming sessions. Instead, if an error is encountered, the streaming session simply terminates without yielding any chunks, which is very much unexpected.

## Affected Areas

This is simply an additional delegate method implementation on `StreamingSession` (`urlSession(session:dataTask:didReceive:completionHandler:)`), and it's _mostly_ an internal change that's transparent to users.

Users should already be handling `OpenAIError` anyway, so hopefully this shouldn't cause any unexpected behavior for anyone. Personally, I would not consider this a breaking change.


SplittyDev:
Btw, as far as I can tell this is different from what's been done in #266, as this actually handles HTTP-specific errors. This is basically what I discussed in #261, which was not resolved by #266.

nezhyborets (approved):
Great, thanks!

PR #288, opened 2025-03-16
Support OpenRouter reasoning responses

SplittyDev:
## What

This PR adds support for the `reasoning` field used by OpenRouter.

In order to unify this with the existing `reasoningContent` field without introducing a breaking API change, I've taken the liberty to make some internal API changes.

- Renamed the existing `reasoningContent` to `deepseekReasoningContent`
- Introduced the new OpenRouter `reasoning` field as `openrouterReasoningContent`
- Marked both `deepseekReasoningContent` and `openrouterReasoningContent` as `internal`
- Introduced a new `public var reasoningContent` which returns whichever reasoning field is actually set, if any

This should be fully backwards-compatible, as the existing `reasoningContent` field has been kept, albeit as a property now.

## Why

As far as I know, there isn't any established standard regarding reasoning responses yet, and it seems that different providers are using different fields for this purpose.

The `reasoning_content` field used by DeepSeek has already been implemented before, but OpenRouter uses a field called `reasoning` instead.


nezhyborets:
Hi @SplittyDev , thanks again for the contribution. Sorry, but in parallel, I made a refactoring to `ChatResult.Choice.Message`, and now `ChatQuery`'s `Message` is not used for parsing responses. So I've implemented this change here https://github.com/MacPaw/OpenAI/pull/288

So, I also added `reasoning` and `reasoning_content`. But I added `reasoning` for Gemini, and `reasoning_content` for DeepSeek. For OpenRouter I couldn't find anything about reasoning in their API Reference about response. Do they return that? Anyway, it would get parsed

SplittyDev:
@nezhyborets no worries!

> So I've implemented this change here https://github.com/MacPaw/OpenAI/pull/288
I believe you meant #289? I've left a comment, please check it out. Makes for a cleaner API imo.

> For OpenRouter I couldn't find anything about reasoning in their API Reference about response. Do they return that?
Yes, OpenRouter uses the `reasoning` field.

PR #289, opened 2025-03-20
Add reasoning and reasoningContent to Message

nezhyborets:
## What

Add `reasoning` and `reasoningContent` to `ChatResult.Choice.Message` to enable parsing of reasoning fields of Gemini and DeepSeek responses

## Why

Requested by users in issues


SplittyDev:
@nezhyborets I still think it would be good to make these fields private and expose them to the API consumer under a single `reasoning` or `reasoningContent` field, but not both, like I did in #288.

Some APIs return `reasoning` and some return `reasoningContent`, but none of them return _both_. Imo it doesn't make much sense that the user has to check both fields manually.

It would be even cleaner to have a custom `Decodable` implementation that tries decoding both `reasoning` and `reasoning_content` keys, and only sets one field in the end. But that's definitely more work.

Krivoblotsky (approved):

gtokman:
this is a response from Groq and i think this will work with it:

```
{
  "id": "chatcmpl-260011a1-d778-45bf-877c-46291ddeb5c3",
  "object": "chat.completion",
  "created": 1741629033,
  "model": "qwen-qwq-32b",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "reasoning": "<>",
        "tool_calls": [
          {
            "id": "call_zgfy",
            "type": "function",
            "function": {
              "name": "get_linked_accounts",
              "arguments": "{}"
            }
          }
        ]
      },
      "logprobs": null,
      "finish_reason": "tool_calls"
    }
  ],
  "usage": {
    "queue_time": 0.21565806099999998,
    "prompt_tokens": 282,
    "prompt_time": 0.019009824,
    "completion_tokens": 181,
    "completion_time": 0.418294019,
    "total_tokens": 463,
    "total_time": 0.437303843
  },
  "system_fingerprint": "fp_c81bd6c71c",
  "x_groq": {
    "id": "req_01jp0jw2qwf6ar1kf7jbza1m7h"
  }
}
```

PR #301, opened 2025-03-31
Added support for audio-preview models, including updates to query pa…

rkvanadea:
Added support for audio-preview models, including updates to query parameters, message parameters, and audio delta.

## What

Add support for gpt-4o-audio-preview model in https://api.openai.com/v1/chat/completions. This allow send and receive audio data directly from model. 

## Why

Good alternative to STT->CHAT COMPLETIONS->TTS or Realtime API.

## Affected Areas

Add model, audio delta, user and assistant messages params.


nezhyborets (approved):
Great, thanks 🤩

PR #303, opened 2025-03-31
Audio stream fix.

rkvanadea:
## What

Added "gpt-4o-mini-audio-preview".
Fix "expiresAt" parameter of audio delta.
Fix StreamingSession onComplete calling - onComplete called before all chunks produced.

## Why

"expiresAt" is timestamp and not String.
Especially for audio chunks chat stream continuation finished before last chunks was yield. urlSession didCompleteWithError func called before last chunks or chunk is parsed by interpreter.

## Affected Areas

GPT models, ChatStreamResult, chat streaming helpers.


nezhyborets (approved):

nezhyborets:
@rkvanadea thanks for the PR. The issue was in the order of execution. Adjusted the code a bit and wrote a test

nezhyborets:
@rkvanadea please let me know if it works for you, you'd have to use `main` instead of a version to try it out

nezhyborets:
@rkvanadea nevermind, I found another issue with it, will be another PR

PR #279, opened 2025-03-07
feat:Support reasoningContent in StreamResult

LimChihi:
## What

Add `reasoningContent` to `ChatStreamResult.Choice.ChoiceDelta`

## Why

Support for the DeepSeek-R1 reasoning content response.

#271 

https://api-docs.deepseek.com/guides/reasoning_model

## Affected Areas

Incremental modification, nil if no return

PR #311, opened 2025-04-03
Support `reasoning` in ChatStreamResult deltas

SplittyDev:
## What

Currently, streaming deltas only support the `reasoning_content` field. But providers such as OpenRouter (and possibly Gemini in OpenAI compatibility mode) use the `reasoning` field instead.

## Why

Support for the `reasoning` field is missing in streaming deltas.

- Support for `reasoning_content` in streaming deltas was added:
  #279
- Support for `reasoning` and `reasoning_content` in regular chat responses was added:
  #289

## Affected Areas

No breaking changes, just a new field in `ChatStreamResult.Choice.ChoiceDelta`.

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatStreamResult.swift
+++ b/Sources/OpenAI/Public/Models/ChatStreamResult.swift
@@ -18,12 +18,25 @@ public struct ChatStreamResult: Codable, Equatable, Sendable {
             public let content: String?
             /// If the audio output modality is requested, this object contains data about the audio response from the model.
             public let audio: ChoiceDeltaAudio?
-            /// The reasoning content of the chunk message.
-            /// Only some model are supported, like DeepSeek-R1.
-            public let reasoningContent: String?
+
             /// The role of the author of this message.
             public let role: Self.Role?
             public let toolCalls: [Self.ChoiceDeltaToolCall]?
+
+            /// Value for `reasoning` field in response.
+            ///
+            /// Provided by:
+            /// - Gemini (in OpenAI compatibility mode)
+            ///   https://github.com/MacPaw/OpenAI/issues/283#issuecomment-2711396735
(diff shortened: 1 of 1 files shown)


SplittyDev:
I have a potential follow-up to this where `reasoning` and `reasoning_content` is unified to streamline the public API. See https://github.com/FiveSheepCo/OpenAI-OpenRouter/commit/46291316914510cac38b470281abc3a10cd8aa87

nezhyborets (approved):
🔥

nezhyborets:
As for combining the fields... I am a bit worries about the following case

Sometime ago I remember someone said that they kinda forward outputs from one model to another. Or they somehow combine the outputs. In this context, I worry that if we combine the fields, the simple decode/encode pair wouldn't work anymore, like:

```swift
let responseData: Data
let resultStruct = JSONDecoder().decode(SomeResultType.self)
let data = JSONEncoder().encode(resultStruct)
```

This could put a reasoning value under a wrong key

SplittyDev:
@nezhyborets I think that's not an issue with the way I am doing the unification in my branch.

I'm doing something like this:

```swift
// ...

internal let _reasoning: String?
internal let _reasoningContent: String?

/// Reasoning content.
public var reasoning: String? {
    _reasoning ?? _reasoningContent
}

// ...

public enum CodingKeys: String, CodingKey {
    // ...
    case _reasoning = "reasoning"
    case _reasoningContent = "reasoning_content"
    // ...
}
```

So the unification of the fields only happens in the public API. When re-encoding the struct, both the original `reasoning` as well as the original `reasoning_content` fields will be preserved.

nezhyborets:
I've been thinking about something like this, and it does seem like it would work! Would gladly make such an addition to the codebase of this repo

PR #312, opened 2025-04-03
Add `error` finish reason

SplittyDev:
## What

This PR adds an additional `error` finish reason.

## Why

Some inference providers, such as OpenRouter, include an `error` finish reason.

Docs: https://openrouter.ai/docs/api-reference/overview#finish-reason

## Affected Areas

`FinishReason`

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatResult.swift
+++ b/Sources/OpenAI/Public/Models/ChatResult.swift
@@ -158,6 +158,7 @@ public struct ChatResult: Codable, Equatable, Sendable {
             case toolCalls = "tool_calls"
             case contentFilter = "content_filter"
             case functionCall = "function_call"
+            case error
         }
     }
 


SplittyDev:
Maybe the `finish_reason` should also contain an `unknown` variant to handle previous unhandled cases.

Looking at Google Cloud Generative AI docs, they have even more finish reasons, such as:
- `finish_reason_unspecified`
- `recitation`
- `spii`
- `prohibited_content`
- `safety`
- `blocklist`
- `other`

All of these would currently cause a `DecodingError/dataCorrupted` error to be thrown, and since every provider seems to be doing their own thing regarding finish reasons, they should probably be handled better.

nezhyborets (approved):

nezhyborets:
🤔🤔🤔
Yes, it would be probably easier to just add something like `.other(String)` because it seems like it would be too verbose to handle all the cases of all the providers, especially since just Google has so many of them. Do you think it's better to do as part of this PR or separately?

SplittyDev:
> 🤔🤔🤔 Yes, it would be probably easier to just add something like `.other(String)` because it seems like it would be too verbose to handle all the cases of all the providers, especially since just Google has so many of them. Do you think it's better to do as part of this PR or separately?

I'd probably rather do it separately. Probably a bit of discussion to be had there, because as far as I can tell, Google literally has "other" as a reason, so we might choose a unique name for the catch-all case. Not too sure about that yet, so if it's not a problem I'd like to merge this first, because it already solves the OpenRouter issue.

nezhyborets:
Sure, done, thanks!

PR #313, opened 2025-04-03
Unify `reasoning` and `reasoningContent` fields

SplittyDev:
This simplifies the API and hides an unnecessary implementation detail.

## What

This PR unifies the `reasoning` and `reasoning_content` fields into a single public `reasoning` getter.

The internal `reasoning` and `reasoningContent` fields are preserved for en- and decoding purposes, to preserve ABI-compatibility and ensure that re-encoding is lossless.

## Why

As discussed in https://github.com/MacPaw/OpenAI/pull/311#issuecomment-2775237198.
There is no reason to publicly expose both, as only one or neither of these fields will be set at a time.

## Affected Areas

`ChatResult`, `ChatStreamResult`


nezhyborets (approved):
Cool!

PR #314, opened 2025-04-04
Support dynamic json schema response format

dounan:
## What

Support a more flexible and dynamic way to define the `response_format.json_schema` parameter. This approach allows the user to also specify `description` and `strict`, as well as pass any `Encodable & Sendable` as the JSON schema itself.

## Why

This unlocks the ability for users to dynamically build up a JSON schema at runtime, vs having to define a `StructuredOutput` at compile time.

The user can choose to define their JSON schema using simple Dictionaries, or specify their JSON schema with a library like https://github.com/kevinhermawan/swift-json-schema.

## Affected Areas

ChatQuery

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatQuery.swift
+++ b/Sources/OpenAI/Public/Models/ChatQuery.swift
@@ -1,6 +1,6 @@
 //
 //  ChatQuery.swift
-//  
+//
 //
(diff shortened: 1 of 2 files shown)


nezhyborets (approved):
Looks great and useful, thanks for the contribution!

PR #316, opened 2025-04-12
Rewrite the parser by more closely following spec

nezhyborets:
## What

Rewrite Server Sent Events parser

## Why

Our parser doesn't follow the spec closely which may lead to bugs like in the connected issue

## Affected Areas

Streaming Interpreter

### DIFF
--- a/Sources/OpenAI/Private/Streaming/ServerSentEventsStreamInterpreter.swift
+++ b/Sources/OpenAI/Private/Streaming/ServerSentEventsStreamInterpreter.swift
@@ -12,6 +12,7 @@ import Foundation
 ///
 /// - Note: This class is NOT thread safe. It is a caller's responsibility to call all the methods in a thread-safe manner.
 final class ServerSentEventsStreamInterpreter <ResultType: Codable & Sendable>: @unchecked Sendable, StreamInterpreter {
+    private let parser = ServerSentEventsStreamParser()
     private let streamingCompletionMarker = "[DONE]"
     private var previousChunkBuffer = ""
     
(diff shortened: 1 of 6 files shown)


nezhyborets:
@JackuXL could you try to switch package to this branch to see if it works for you?

JackuXL:
> @JackuXL could you try to switch package to this branch to see if it works for you?
> 
> 315-streaming-session-is-ended-before-any-output

I have tried, but it still does not work. The output is the same as before.

nezhyborets:
Omg, crazy. Thanks for trying. Will try to find how to configure Volcano on my side to reproduce

nezhyborets:
It could be very helpful if you'd put a breakpoint in StreamingSession and print the raw response data. It returns data so you'd have to print `String(data: data, encoding: .utf8)`. The response may come multiple times, the first one may be enough

nezhyborets:
On line 63 should work 

image

nezhyborets:
Is it the first and the only response in streaming session? If so, then there would basically be nothing to parse. I am not sure why a platform would respond in such a manner 🤔 interesting

nezhyborets:
Can you print the query?

JackuXL:
> Is it the first and the only response in streaming session? If so, then there would basically be nothing to parse. I am not sure why a platform would respond in such a manner 🤔 interesting 

The platform enables developers to bring websearch to any LLM, and use it through standard OpenAI API. 

Later I would try to stream using official OpenAI Python SDK and see if it could work.

nezhyborets:
It would be great, thanks!

JackuXL:
It seems that the latest version works, and the previous error is due to volcengine itself. Thanks!

nezhyborets:
Yay! Thanks for checking!

Krivoblotsky (approved):

nezhyborets:
Hi @JackuXL, it turned out that there had been multiple issues with the new parser. I've release the fixes in 0.4.1. So maybe it will work better for you too

JackuXL:
> Hi @JackuXL, it turned out that there had been multiple issues with the new parser. I've release the fixes in 0.4.1. So maybe it will work better for you too

Ok, thanks for the reminder!

PR #354, opened 2025-06-23
feat: generate latest Types based on latest openai openapi yaml

atom2ueki:
## What

re-generate Components.swift based on latest openai-openapi.yaml
https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml

## Why

I will raise another PR adding mcp support later, just wants to separate the pr easier to review

## Affected Areas

Nothing affected, test cases pass

However, I found that test cases require iOS 16 but package.swift minimum support iOS 13, when I run test cases locally need to bump up the version to 16 first

## Extra

Now the main blocker to keep it up to date is the unstable openai yaml, their yaml file not always works well, buggy, maybe we can having a workflow help us automate maintain it, just need add `swift-openapi-generator` dependency at package.swift

### DIFF
--- a/Sources/OpenAI/Public/Schemas/Generated/Components.swift
+++ b/Sources/OpenAI/Public/Schemas/Generated/Components.swift
@@ -19,12 +19,18 @@ import struct Foundation.Date
 /**
  The code here is generated using OpenAPI document provided by OpenAI.
  
- # OpenAI OpenAPI commit: #
- 498c71dd Update openapi.yaml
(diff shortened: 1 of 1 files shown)


nezhyborets:
Hi @atom2ueki , as you gave me green flag to proceed, I've implemented all the new fields in Responses API, both streaming and not. Now in process of testing and debugging. I'm not sure we should do it in parallel. But we could take a look at your issue with testing in this PR. Could you push the change you make to run tests so that I could take a look?

PR #355, opened 2025-06-23
Update Responses API with new tools

### LINKED ISSUE #347: openai-openapi code generation issue
The maintenance of [openai-openapi](https://github.com/openai/openai-openapi) is not very active, and since we rely on openapi to generate code, many new features cannot be provided in a timely manner, such as image generation by gpt1, support for MCP, and so on, in the [Responses API](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools).

nezhyborets:
OpenAI introduced new tools like Image Generation, MCP etc. This PR implements that


atom2ueki:
@nezhyborets LGTM, my PR missing main MCPTool struct, closed
question: why we need keep those Edited Schemas, its a bit hard to maintain actually

nezhyborets:
They definitely are hard to support, and in this PR I got rid of one (I think). 

Today I've started to add comments above "Edited" types to explain the reason, because sometimes I forget why I did that.

But most of the time it's because some generated types would have a bad interface. I.e. it would be composed of other types, and the fields of those types would be "value1" "value2" etc. It's very far from user friendliness. But we also can't "prettify" all such types because then we would rewrite everything that is generated. 

I'm not sure yet how to be about it, I guess we'll figure out in the process. If you think some edited type can be removed because it doesn't provide much value - please let me know. I too would like to keep them to minimum

atom2ueki:
Thanks, I will go ahead implement MCP based on this

Krivoblotsky (approved):
Nice!

PR #356, opened 2025-06-25
Support OpenRouter models endpoint

SplittyDev:
## What

This PR makes the `/models` endpoint work with OpenRouter.

## Why

Unlike OpenAI, OpenRouter doesn't use the `object` and `owned_by` fields.

I've chosen to use reasonable defaults for `object` (i.e. `list` in `ModelsResult` and `model` in `ModelResult`), and an empty string for `owned_by`, since I don't think there's a better default for that one.

The `ModelsResult` and `ModelResult` structs themselves aren't changed, and this PR makes use of the existing `parsingOptions` mechanism to supply the default values.

## Affected Areas

- Decoding of `ModelsResult` and `ModelResult`, if relaxed parsing options are specified
- No breaking changes


nezhyborets (approved):
Great, thanks!

Issue #361, opened 2025-07-01
Structured Output Configuration


baskurthalit:
Hi,

In responseFormat of ChatQuery, I want structured output with .jsonSchema. But its associated value of **StructuredOutputConfigurationOptions** is not accessible outside of module like other cases.

<img width="550" alt="Image" src="https://github.com/user-attachments/assets/2d07b46e-564f-4ab0-8e7c-c66d32d3a451" />

nezhyborets:
Hi @baskurthalit , thanks for the request! I've been working on it just now as part of https://github.com/MacPaw/OpenAI/issues/351, so a fix should be ready soon

nezhyborets:
`.jsonSchema(...)` definition is currently only suitable to build a schema by hand, like this:
```
.jsonSchema(
    .schema(
        .type(.object),
        .properties([
            "title": .schema(
                .type(.string)
            )
        ])
    )
),
```

In your case, could you please try to conform `SummaryResult` to `JSONSchemaConvertible` and use `.derivedJsonSchema` definition? Like this:
```
responseFormat: .jsonSchema(
    .init(
        name: "jsonSchema",
        description: nil,
        schema: .derivedJsonSchema(SummaryResult.Type),
        strict: true
    )
)
```

nezhyborets:
@baskurthalit thank you very much, don't worry about it, we'll see how it goes from now on with these APIs and if they need more improvements

Issue #351, opened 2025-06-10
Add JSON Schema builder for Response API Structured output return format


mireabot:
**Is your feature request related to a problem? Please describe.**
The current version of the package gives support to define custom struct as JSON Schema for chat/completions API. While query builder for /responses API only supports .text as clear type of return

**Describe the solution you'd like**
For chat completions you can build query like this
```swift
let query = ChatQuery(
    messages: [.system(.init(content: "Best Picture winner at the 2011 Oscars"))],
    model: .gpt4_o,
    responseFormat: .derivedJsonSchema(name: "movie-info", type: MovieInfo.self)
)
```

For Response Query I found the workaround accessing OpenAIRuntime, but could be like this

```swift
let practicalQuery = CreateModelResponseQuery(
            input: .textInput("Hi there, I'm having trouble with my computer"),
            model: .gpt4_1,
            text: .derivedJsonSchema(name: "movie-info", type: MovieInfo.self),
            toolChoice: .ToolChoiceOptions(.auto),
            tools: [.webSearchTool(.init(_type: .webSearchPreview20250311))]
        )
```
**Current workaround**
```swift
let practicalQuery = CreateModelResponseQuery(
            input: .textInput("Hi there, I'm having trouble with my computer"),
            model: .gpt4_1,
            instructions: "Analyze the user's message and provide a helpful response",
            maxOutputTokens: 500,
            store: true,
            stream: false,
            text: .init(
                format: .TextResponseFormatJsonSchema(
                    Components.Schemas.TextResponseFormatJsonSchema(
                        _type: .jsonSchema,
                        description: "Structured response for user support",
                        name: "support_response",
                        schema: {
                            var container = OpenAPIRuntime.OpenAPIObjectContainer()
                            container.value["type"] = "object"
                            container.value["properties"] = [
                                "response": ["type": "string"],
                                "category": ["type": "string", "enum": ["technical", "general", "urgent"]],
                                "confidence": ["type": "number", "minimum": 0, "maximum": 1],
                                "next_steps": [
                                    "type": "array",
                                    "items": ["type": "string"]
                                ]
                            ]
                            container.value["required"] = ["response", "category", "confidence", "next_steps"]
                            container.value["additionalProperties"] = false
                            return Components.Schemas.ResponseFormatJsonSchemaSchema(additionalProperties: container)
                        }(),
                        strict: true
                    )
                )
            ),
            toolChoice: .ToolChoiceOptions(.auto),
[...]

PR #362, opened 2025-07-03
351 add json schema builder for response api structured output return format

nezhyborets:
## What

Unify approach for defining JSONSchemas in Chat Completions and Responses API

Improve JSONSchema definition (make it a bit less wordy)

Update readme: 
* Make examples of all three ways for defining Schema. Use "collapsible" feature of Markdown to hide big examples for easier navigation.
* Make "Structured Outputs" to be its own section at a higher level, because it applies to both Chats and Responses. 
* Move "Structured Outputs" below Audio.

Generally, I am planning on following "Guides" structure of OpenAPI Platform Doc to structure our doc.

Screenshot 2025-07-03 at 17 06 19

## Why

Requests to improve API regarding JSONSchema:
https://github.com/MacPaw/OpenAI/issues/351
https://github.com/MacPaw/OpenAI/issues/361

As for the doc, we've discussed that it should be improved incrementally. This is the first step.


nezhyborets:
Screenshot 2025-07-03 at 17 12 48

PR #377, opened 2025-08-13
Update LinkPreview.swift

mehmetbaykar:
**Fix Xcode 26.0 Swift Concurrency error** 

## What

* Refactor `LinkPreview` to use Swift Concurrency.
* Fetch metadata with `async/await` and set `LPLinkView.metadata` on the main actor:

  * `Task { @MainActor in uiView.metadata = try await fetchMetadata() }`
  * Introduce `fetchMetadata() async throws -> LPLinkMetadata`.

## Why

* In Xcode 26.0 with stricter concurrency checks, the previous completion-handler approach attempted to set `uiView.metadata` (a `@MainActor`-isolated property) from a `@Sendable`/background context, triggering:
  *“Main actor-isolated property 'metadata' cannot be mutated from a Sendable closure.”*
* Using `async/await` removes the sendable-closure issue and guarantees UI updates occur on the main actor, aligning with Swift 6 concurrency rules.

## Affected Areas

* `LinkPreview.swift` only.
* No behavioral change to the UI; thread-safety and compile-time correctness improved.
* Note: Requires async/await (iOS 15+/macOS 12+).

### DIFF
--- a/Demo/DemoChat/Sources/UI/Images/LinkPreview.swift
+++ b/Demo/DemoChat/Sources/UI/Images/LinkPreview.swift
@@ -18,16 +18,14 @@ struct LinkPreview: UIViewRepresentable {
     }
     
     func updateUIView(_ uiView: UIViewType, context: Context) {
-        LPMetadataProvider().startFetchingMetadata(for: previewURL) { metadata, error in
-            if let error = error {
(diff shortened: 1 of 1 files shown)


nezhyborets (approved):
Great, thanks!

PR #380, opened 2025-08-18
Make ChoiceDeltaToolCall.index optional

batanus:
## What

Make ChoiceDeltaToolCall.index optional

## Why
According to the OpenAI documentation, this property is not required in the response. ChatGPT always returns the index property, but some other providers omit it.


JaredConover (approved):
I noticed this is an issue for other endpoints/properties as well.
The openAI spec doesn't mark very many of these params as required, so I don't think parsing should automatically fail if they're not present.

nezhyborets (approved):

PR #382, opened 2025-09-03
Make response props parsing more robust for OpenAI compatible APIs

zats:
I noticed that local open-ai compatible response would fail parsing oftentimes with some fields missing;
Made toolChoice and the token metrics optional to satisfy.

I can see how it might be a breaking change for some consumers but unsure if it's better to provide a default value (might also break assumptions)


nezhyborets:
Hi @zats , thanks for the PR. The `Components` is a file that is generated. I've regenerated it, not sure if optionality of those fields changed. And as for `ResponseObject` - a change has been merged that makes `toolChoice` optional

PR #387, opened 2025-09-24
Fix: Consider event type in data payload when processing model response streams

JaredConover:
## What
Enable the parsing of `type` property inside of the event.data object for mapping model responses.

## Why
Parsing of streamed events from /responses endpoint only looks at the type property of the SSE event itself, and not at the type property inside the events data payload. As a result it is not possible to parse properly formed events.

When attempting to stream events from the /responses api using OpenAI.responses.createResponseStreaming() it seems there is a bug that prevents the events from being processed.

Despite receiving properly formatted events, ie:
```
data: {"type":"response.output_text.delta","sequence_number":0,"delta":"#", ...}
```

ModelResponseEventsStreamInterpreter.processEvent for:
```
Printing description of event:
▿ Event
  - id : nil
  ▿ data : 69 bytes
    - count : 69
    ▿ pointer : 0x000000011450ec90
      - pointerValue : 4635815056
  - decodedData : "{\"type\":\"response.output_text.delta\",\"sequence_number\":0,\"delta\":\"#\"}"
  - eventType : "message"
  - retry : nil
```

Produces error:
```
Printing description of error:
▿ InterpreterError
  - unknownEventType : "message"
```

The "message" event type is not present in the event received and seems to come from line 207 of the `ServerSentEventsStreamParser` that sets a default event type (for the sse event) to "message" by default if there is no value in the event *field* of the sse event.

I believe the bug is that `ModelResponseEventsStreamInterpreter.processEvent` refers to this default value for the SSE event field when attempting to parse a `ModelResponseStreamEventType` when it should be using the `type` property inside the SSE event's data object as per the openAI spec. We can see is "response.output_text.delta" in the example above.

## Affected Areas

This updates how streamed model responses are parsed and processed.

### DIFF
--- a/Sources/OpenAI/Private/Streaming/ModelResponseEventsStreamInterpreter.swift
+++ b/Sources/OpenAI/Private/Streaming/ModelResponseEventsStreamInterpreter.swift
@@ -55,22 +55,30 @@ final class ModelResponseEventsStreamInterpreter: @unchecked Sendable, StreamInt
     }
     
     private func processEvent(_ event: ServerSentEventsStreamParser.Event) throws {
-        var finalEvent = event
-        if event.eventType == "response.output_text.annotation.added" {
-            // Remove when they have fixed (unified)!
-            //
-            // By looking at [API Reference](https://platform.openai.com/docs/api-reference/responses-streaming/response/output_text_annotation/added)
-            // and generated type `Schemas.ResponseOutputTextAnnotationAddedEvent`
-            // We can see that "output_text.annotation" is incorrect, whereas output_text_annotation is the correct one
-            let fixedDataString = event.decodedData.replacingOccurrences(of: "response.output_text.annotation.added", with: "response.output_text_annotation.added")
-            finalEvent = .init(id: event.id, data: fixedDataString.data(using: .utf8) ?? event.data, decodedData: fixedDataString, eventType: "response.output_text_annotation.added", retry: event.retry)
+        var eventType = event.eventType
+
+        if eventType == "message" { // This is currently the default if no SSE event name is specified
+            struct _TypeEnvelope: Decodable { let type: String }
+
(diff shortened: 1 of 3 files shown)


nezhyborets:
BTW I don't know why but it doesn't happen for me, can't reproduce

Sources/OpenAI/Private/Streaming/ModelResponseEventsStreamInterpreter.swift
```diff
@@ -55,25 +55,24 @@ final class ModelResponseEventsStreamInterpreter: @unchecked Sendable, StreamInt
...
-            // By looking at [API Reference](https://platform.openai.com/docs/api-reference/responses-streaming/response/output_text_annotation/added)
-            // and generated type `Schemas.ResponseOutputTextAnnotationAddedEvent`
-            // We can see that "output_text.annotation" is incorrect, whereas output_text_annotation is the correct one
-            let fixedDataString = event.decodedData.replacingOccurrences(of: "response.output_text.annotation.added", with: "response.output_text_annotation.added")
-            finalEvent = .init(id: event.id, data: fixedDataString.data(using: .utf8) ?? event.data, decodedData: fixedDataString, eventType: "response.output_text_annotation.added", retry: event.retry)
+        let finalEvent = event.fixMappingError()
+        var eventType = finalEvent.eventType
+
+        /// If the SSE `event` property is not specified by the provider service, our parser defaults it to "message" which is not a valid model response type.
+    /// In this case we check the `data.type` property for a valid model response type.
```
JaredConover, line 62:
```suggestion
        /// If the SSE `event` property is not specified by the provider service, our parser defaults it to "message" which is not a valid model response type.
        /// In this case we check the `data.type` property for a valid model response type.
```

nezhyborets (approved):

PR #399, opened 2025-12-12
Add annotations to chat stream result

batanus:
## What
Add Annotations property to ChatStreamResult object

## Why
When OpenAI uses web tool call in stream mode - it returns annotation object for ChatStreamResult as well

## Affected Areas
`ChatStreamResult`, `Annotation` structs


nezhyborets (approved):
Thanks!

PR #402, opened 2026-01-01
Add reasoningContent for model chat

gongxh13:
## What

Add reasoningContent to the model invocation request body.

## Why
Some models require the **reasoningContent** parameter to be passed in multiple calls within the same conversation round; otherwise, an error will occur. A typical example is the **DeepSeek Reasoner** model. See the official documentation: [https://api-docs.deepseek.com/guides/thinking_mode#tool-calls](https://api-docs.deepseek.com/guides/thinking_mode#tool-calls)

## Affected Areas

Add a **reasoningContent** field to the message, defaulting to **nil**. It is passed to the model invocation only when explicitly provided.

### DIFF
--- a/Sources/OpenAI/Public/Models/ChatQuery.swift
+++ b/Sources/OpenAI/Public/Models/ChatQuery.swift
@@ -292,6 +292,7 @@ public struct ChatQuery: Equatable, Codable, Streamable, Sendable {
         public init?(
             role: Role,
             content: String? = nil,
+            reasoningContent: String? = nil,
             imageData: Data? = nil,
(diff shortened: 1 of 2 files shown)


nezhyborets:
Hi @gongxh13 , thanks for the contribution!

I wonder if we should somehow visually separate the fields that are present in OpenAI API Reference and those that are related to some other providers. Like maybe put these fields at the end of their lists, write a comment... But for now I'm just going to merge and maybe return to this question later

nezhyborets (approved):

ghost:
谢谢

PR #408, opened 2026-04-18
feat: add refusal field to streaming chat completion delta

### LINKED ISSUE #249: Improve Structured Outputs Support: Add refusal field to Chat Completion Object
**Is your feature request related to a problem? Please describe.**
Structured Outputs have been implemented thanks to @andgordio. With Structured Outputs they've added additional field for when model is unable to generate response in a specified format: `refusal`, see [Refusals in Structured Outputs Guide](https://platform.openai.com/docs/guides/structured-outputs/refusals#refusals). For more info.

The field appears in both streamable and regular chat completion objects:
1. Chat Completion Object - choices.message.refusal [[API reference](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices)]
2. Chat Completion Chunk Object - choices.delta.refusal [API Reference](https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices)

**Describe the solution you'd like**
Simply add the field to our structures that represent `choices.message` and `choices.delta`. It might require refactoring though. Currently we use `ChatCompletionMessageParam` to both create chat completion and parse Chat Completion Object in a response. But looking at the API reference, it seems that request body and response format are different and so different Swift structured may be used for clearer distinction. Also, the separation into different types makes sense for a request, but not so much for response: API Reference does not specify such distinction, and I don't think we should either.

**Describe alternatives you've considered**
An alternative would be to continue using `ChatCompletionMessageParam` to decode Chat Completion Object, it may be faster to implement this way.

**Additional context**

mvanhorn:
## Summary

- Adds `refusal: String?` to `ChatStreamResult.Choice.ChoiceDelta`, matching the existing field in `ChatResult.Choice.Message`
- Adds `refusal: [ChatCompletionTokenLogprob]?` to `ChatStreamResult.Choice.ChoiceLogprobs`, matching the non-streaming logprobs
- Adds `refusal` to the `CodingKeys` enum for correct JSON deserialization
- `swift build` passes cleanly

Per the [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices), `choices.delta.refusal` is returned when the model refuses to generate a response in the requested Structured Output format.

Closes #249

## Test plan

- [x] `swift build` succeeds
- [ ] Verify streaming responses with Structured Outputs that trigger a refusal now populate the `refusal` field


nezhyborets (approved):

nezhyborets:
Thanks!

mvanhorn:
Much appreciated, @nezhyborets. The refusal field on streaming deltas brings the client up to parity with the API.

PR #428, opened 2026-06-04
Document OpenAI-compatible base URL configuration

cerebrixos:
This keeps the default OpenAI endpoint unchanged while documenting how to configure another OpenAI-compatible endpoint through OPENAI_BASE_URL.\n\nWhy this may be useful:\n- local inference servers\n- private model gateways\n- self-hosted routers\n- observability and enterprise control planes\n\nThe example uses the existing Configuration fields and does not add a provider-specific integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,22 @@ This SDK is more focused on working with OpenAI Platform, but also works with ot
 
 Use `.relaxed` parsing option on Configuration, or see more details on the topic [here](#support-for-other-providers)
 
+If you need to switch between OpenAI and another OpenAI-compatible endpoint in different environments, you can build the configuration from `OPENAI_BASE_URL`:
+
+```swift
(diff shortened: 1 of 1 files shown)


Krivoblotsky (approved):

nezhyborets:
Hi @cerebrixos, thanks for the contribution!
Do you mean you want to show that URL string and the token could be retrieve from `ProcessInfo.processInfo.environment` before passing to `Configuration.init`?

cerebrixos:
Yes exactly — the idea is to show that Configuration already supports a custom base URL, so users can point it at any OpenAI-compatible endpoint by reading OPENAI_BASE_URL from the environment (or hardcoding it). No new API surface needed — just a documentation example showing the existing host parameter in Configuration.init can be set to a non-OpenAI URL. Happy to adjust the example to explicitly use ProcessInfo.processInfo.environment["OPENAI_BASE_URL"] if that makes the intent clearer.

PR #429, opened 2026-06-09
feat(demo): provider picker for OpenAI-compatible hosts

### LINKED ISSUE #308: Make Demo configurable for providers through UI
**Is your feature request related to a problem? Please describe.**
Currently we can only set API Key and the demo is hardcoded to work with OpenAI

**Describe the solution you'd like**
I think we can put more UI into the view that holds API Key field. I think we should have predefined sets of values, like if user selects `Gemini` using some new picker - `host` field would get auto populated with `generativelanguage.googleapis.com`. If a user wants to put something custom, they'd just edit the fields manually or select `Custom...` in the same picker. I believe we should also store whatever is inputed, for the following launches.

**Describe alternatives you've considered**
I'm not sure if there are other alternatives besides doing it in UI. Launch arguments? They're not passable on iOS

**Additional context**
The issue rose up during debugging of this issue https://github.com/MacPaw/OpenAI/issues/283. But I believe such functionality would be helpful for other usages

tsushanth:
## Why

Closes #308.

The Demo is hardcoded to \`api.openai.com\`, so testing against any other OpenAI-compatible endpoint (Gemini's compat layer, Groq, OpenRouter, a local server, etc.) requires editing source. The issue specifically asks for a predefined provider picker that auto-populates host/basePath, with a \"Custom…\" escape hatch for everything else, and persistence across launches.

## What

- New \`APIProvider\` enum (\`Demo/App/APIProvider.swift\`) with cases for OpenAI, Gemini, Groq, OpenRouter, and Custom. Each known case carries its default host + basePath.
- \`APIKeyModalView\` gains a provider Picker. Host/basePath text fields are read-only for known providers and editable when \`Custom…\` is selected.
- \`provider\`, \`host\`, \`basePath\`, and \`apiKey\` are all \`@AppStorage\` keys, so the chosen configuration survives relaunch.
- \`APIProvidedView\` builds \`OpenAI.Configuration\` from the stored values and rewires all stores on any change.

## Screenshots / behaviour

- Open the Demo, pick \`Gemini (OpenAI-compatible)\`: host becomes \`generativelanguage.googleapis.com\`, basePath becomes \`/v1beta/openai\`.
- Pick \`Custom…\`: both fields go editable so you can point at \`localhost:8080\` or any internal endpoint.
- Continue button is gated on a non-empty API key + non-empty host.

## Build

\`xcodebuild -project Demo/Demo.xcodeproj -scheme Demo -destination 'generic/platform=iOS Simulator' build\` → \`** BUILD SUCCEEDED **\`.

## Scope

SDK code is untouched — this is purely a Demo improvement, as the issue requested. Happy to follow up with a similar picker in the iOS example or take docs PRs separately.

### DIFF
--- a/Demo/App/APIKeyModalView.swift
+++ b/Demo/App/APIKeyModalView.swift
@@ -13,14 +13,35 @@ struct APIKeyModalView: View {
     let isMandatory: Bool
 
     @Binding private var apiKey: String
+    @Binding private var providerRaw: String
+    @Binding private var host: String
+    @Binding private var basePath: String
+
     @State private var internalAPIKey: String
+    @State private var internalProvider: APIProvider
+    @State private var internalHost: String
+    @State private var internalBasePath: String
 
     public init(
         apiKey: Binding<String>,
+        providerRaw: Binding<String>,
+        host: Binding<String>,
+        basePath: Binding<String>,
         isMandatory: Bool = true
     ) {
         self._apiKey = apiKey
+        self._providerRaw = providerRaw
+        self._host = host
+        self._basePath = basePath
         self._internalAPIKey = State(initialValue: apiKey.wrappedValue)
+        let resolvedProvider = APIProvider(rawValue: providerRaw.wrappedValue) ?? .openAI
+        self._internalProvider = State(initialValue: resolvedProvider)
+        self._internalHost = State(initialValue: host.wrappedValue.isEmpty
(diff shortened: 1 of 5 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #447, opened 2026-09-18
fix: decode failed responses carrying unrecognized error codes

sachaservan:
`ResponseErrorCode` was generated as a `@frozen` enum from the spec's `enum` list, so a `response.failed` event with any other `code` threw a `DecodingError` and the human-readable message never reached the caller. OpenAI adds codes without a spec bump, and OpenAI-compatible servers emit their own (the Tinfoil router now sends `upstream_error`, `server_is_overloaded`, `model_unavailable` — tinfoilsh/confidential-model-router#484).

The code is now an open `String`. The generation pipeline gains `Scripts/open_response_error_code.py`, which strips the enum list from the prepared spec so `make generate` keeps producing the open type. The spec's known values are exposed as constants on `Components.Schemas.ResponseErrorCodes`.

`swift package diagnose-api-breaking-changes 0.5.1` reports the same three pre-existing breaks on `main` and this branch; the enum→typealias change is not flagged, so it is called out in the changelog instead.


sachaservan:
Opened against the wrong repository by mistake; this targets the tinfoilsh fork.

